PR #1, opened 2021-07-23
AMP Ads Integration

sergunich:
What's done:
- Fetching of ampConfig and adsConfig from EP and storing them in DB
- Implemented AMP "sanitizer" to inject Ads before optimizations from AMP plugging started
- Support for multiple selectors and placing limits
- Support of ad-blocking rules
- Injection of targeting information in ad components
- Support for different `relative` options

Note: this PR intentionally doesn't include `vendor/` dir to ease the review - I will commit dependencies at the end

### DIFF
--- a/Empire/AdminSettings.php
+++ b/Empire/AdminSettings.php
@@ -131,9 +131,17 @@ class AdminSettings {
                 <p><label>% of ads on Empire: <input type="text" name="empire_percent" id="empire_percent" value="<?php echo $empire_test; ?>" /></label></p>
                 <p><label>Key-Value for Split Test: <input type="text" name="empire_value" id="empire_value" value="<?php echo $empire_value; ?>" /></label></p>
 
-                <p><label><input type="checkbox" name="empire_amp_ads_enabled"
-                                 id="empire_amp_ads_enabled" <?php echo $amp_ads_enabled ? 'checked' : ''; ?>> AMP Ads
-                        Enabled</label></p>
+                <p>
+                  <label>
+                    <input
+                      type="checkbox"
+                      name="empire_amp_ads_enabled"
+                      id="empire_amp_ads_enabled"
+                      <?php echo $amp_ads_enabled ? 'checked' : ''; ?>
+                    />
+                    AMP Ads Enabled
+                  </label>
+                </p>
 
                 <p><input type="submit" value="Update"/></p>
             </form>
--- a/Empire/AdminSettings.php
+++ b/Empire/AdminSettings.php
@@ -43,6 +43,7 @@ class AdminSettings {
                 update_option( 'empire::one_trust_id', $_POST['empire_one_trust_id'] ?: '', false );
                 update_option( 'empire::sdk_key', $_POST['empire_sdk_key'] ?: '', false );
                 update_option( 'empire::site_id', $_POST['empire_site_id'] ?: '', false );
+                update_option( 'empire::amp_ads_enabled', isset( $_POST['empire_amp_ads_enabled'] ) ? true : false, false );
                 $this->empire->sdk->updateToken( $_POST['empire_sdk_key'] );
 
                 echo '<h3>Updates Saved</h3>';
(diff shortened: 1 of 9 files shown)


Empire/AdminSettings.php
```diff
@@ -129,6 +131,10 @@ public function showSettings() {
                 <p><label>% of ads on Empire: <input type="text" name="empire_percent" id="empire_percent" value="<?php echo $empire_test; ?>" /></label></p>
                 <p><label>Key-Value for Split Test: <input type="text" name="empire_value" id="empire_value" value="<?php echo $empire_value; ?>" /></label></p>
 
+                <p><label><input type="checkbox" name="empire_amp_ads_enabled"
+                                 id="empire_amp_ads_enabled" <?php echo $amp_ads_enabled ? 'checked' : ''; ?>> AMP Ads
+                        Enabled</label></p>
```
ZLevine, line 136: The tabbing here should probably be cleaned up 👍  If it's a line length issue on the `<input />` itself, you could always put the attributes on new lines like we do with JSX:

```php
<input
  type="checkbox"
  name="empire_amp_ads_enabled"
  id="empire_amp_ads_enabled"
  <?php echo $amp_ads_enabled ? 'checked' : ''; ?>
>
```

Also, this input is missing the trailing `/>`; we always include them if this is self-closing

Empire/AmpAdsInjector.php
```diff
@@ -0,0 +1,102 @@
...
+                continue;
+            }
+
+            $url = $targeting['url'];
+            $components = [];
+            switch ($rule['component']) {
+                case 'PATH':
+                    $components = [parse_url($url)['path'] ?? '/'];
+                    break;
+                case 'URL':
```
ZLevine, line 65: We don't expose the `URL` component in the dashboard:

image

Is this something that's programmatically set and returned by the API? If we aren't using it, we should probably remove it.

Empire/SDK/EmpireSdk.php
```diff
@@ -47,7 +47,8 @@ class EmpireSdk {
     public function __construct(
         string $siteGuid,
         ?string $token = null,
-        string $url = 'https://api.empireio.com/graphql'
+        //string $url = 'https://api.empireio.com/graphql'
+        string $url = 'http://api.lcl.empireio.com:8000/graphql'
```
ZLevine, line 51: I think you forgot to revert this one 👍

Empire/SDK/EmpireSdk.php
```diff
@@ -211,7 +212,65 @@ public function queryContentIdMap( $first, $skip ) {
             )
         );
         $result = $this->runQuery( $gql );
-        return $result->data->contentIdMap;
+        return $result['data']['contentIdMap'];
+    }
+
+    public function queryAdConfig() {
+        $gql = ( new Query( 'appAds' ) );
```
ZLevine, line 219: One thing I hadn't considered—the WP plugin lives in its own repo and we always make the GQP API backwards-compatible, but we aren't really testing the plugin when we go to deploy API changes.

We may want to consider putting some safety checks into the build pipeline to test the plugin as well when we make changes to the API. This already implicitly happens on the frontend since the API/frontend are in a monorepo where all our tests run.

What would be _really_ cool is to have a test WP site that pulls the latest plugin and runs tests to make sure ads display etc. as part of the API test pipeline.

ZLevine (commented):
LGTM, but Justin should definitely review all the PHP. I did see one place where we're referencing the local API though.

Empire/SDK/EmpireSdk.php
sergunich: Fixed :+1:

sergunich:
Hey @jdemaris, this PR is overall ready for review. Only a couple of fixes after Zach's review are left to implement.

jdemaris (approved):
This is looking good!

Empire/AmpAdsInjector.php
```diff
@@ -0,0 +1,149 @@
...
+        switch ($relative) {
+            case 'inside_start':
+                return $elem->insertBefore($ad, $elem->firstChild);
+            case 'inside_end':
+                return $elem->appendChild($ad);
+            case 'after':
+                return $elem->parentNode->insertBefore($ad, $elem->nextSibling);
+            case 'before':
+                return $elem->parentNode->insertBefore($ad, $elem);
+            case 'sticky_footer':
```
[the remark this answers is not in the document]
sergunich: Yeah, AMP doesn't have anything that looks like our `sticky_end` component

Empire/SDK/EmpireSdk.php
sergunich: Yeah, great idea! Created task for this https://empireio.atlassian.net/browse/EP-708

PR #5, opened 2021-08-22
EP-571: Build Plugin on Release

jdemaris:


.github/workflows/eks-production-release.yaml
```diff
@@ -0,0 +1,89 @@
...
+          provider: aws # default is linode
+          region: ${{ env.AWS_REGION }}
+          access_key: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          secret_key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
+      # Push zip to S3 Bucket
+      - name: Deploy
+        run: |
+          s3cmd put empire-${{ env.BUILD_NUMBER }}.zip s3://${{ env.AWS_BUCKET }}/
+          s3cmd put empire-${{ env.BUILD_NUMBER }}.zip s3://${{ env.AWS_BUCKET }}/empire-latest.zip
```
sergunich, line 65: Why are we uploading this to S3? What happens next? Do we manually add it to the release?

sergunich (approved):
This is awesome! Thanks for taking care of this!

.github/workflows/eks-production-release.yaml
rrdial: Yea, I think it would be preferable to upload the package directly to the Release itself and not to S3. I found [this action](https://github.com/marketplace/actions/github-action-publish-binaries) that should be able to do to it as part of the build process.

jdemaris: Ah I should add to the release, yes.

I want to put a link to “latest” in the Empire Dashboard and S3 was my initial thought, but Id rather just link to the latest  release.

PR #14, opened 2021-09-16
Upgrade Prebid Version

jdemaris:
Creating this as a draft until the dependencies are deployed in Empire Platform

### DIFF
--- a/Empire/PageInjection.php
+++ b/Empire/PageInjection.php
@@ -209,7 +209,7 @@ class PageInjection {
 
                 (function() {
                     function loadAds() {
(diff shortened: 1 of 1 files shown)


jdemaris:
I have confirmed that current bidding partners work correctly with the latest Prebid.JS version.

jdemaris:
Note that Conversant doesn't show in Headerbid Expert after making this change, but that is because they changed their bid server URL. I have submitted a Pull Request to the Headerbid Expert codebase to address this: https://github.com/prebid/headerbid-expert/pull/38

PR #15, opened 2021-09-24
Export unquoted content titles

devbis:
Wordpress stores titles html quoted in DB. Unquote titles before sending.

### DIFF
--- a/Empire/Empire.php
+++ b/Empire/Empire.php
@@ -466,7 +466,7 @@ class Empire {
         # In order to support non-standard post metadata, we have a filter for each attribute
         $external_id = \apply_filters('empire_post_id', $post->ID);
         $canonical = \apply_filters('empire_post_url', $canonical, $post->ID);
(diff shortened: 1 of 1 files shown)


Empire/Empire.php
```diff
@@ -466,7 +466,7 @@ public function syncPost( $post ) {
         # In order to support non-standard post metadata, we have a filter for each attribute
         $external_id = \apply_filters('empire_post_id', $post->ID);
         $canonical = \apply_filters('empire_post_url', $canonical, $post->ID);
-        $title = \apply_filters('empire_post_title', $post->post_title, $post->ID);
+        $title = htmlspecialchars_decode(\apply_filters('empire_post_title', $post->post_title, $post->ID));
```
jdemaris, line 469: Let's move this decode deeper inside so that the filter still has full control of what is injected
```suggestion
        $title = htmlspecialcars_decode($post->post_title);
        $title = \apply_filters('empire_post_title', $title, $post->ID);
```

jdemaris (approved):
LGTM

PR #16, opened 2021-09-27
EP-440: inline Ad containers and configs into page

sergunich:


Empire/PageInjection.php
```diff
@@ -198,6 +222,12 @@ public function injectPixel() {
                 window.empire.apps = window.empire.apps || {};
                 window.empire.apps.ads = window.empire.apps.ads || {};
                 window.empire.apps.ads.config = window.empire.apps.ads.config || {};
+            <?php if ( $this->empire->useInjectedAdsConfig() ) { ?>
+
```
ZLevine, line 226: You can probably remove this empty line

sergunich: This is intentional; with this empty line, it looks much nicer in output HTML. Without it, there will be weird indentation.

Empire/PrefillAdsInjector.php
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        if ($styles) {
+            $this->injectStyles($contentDom, $styles);
+        }
+
+        return $contentDom->saveHTML();
+    }
+
+    public function injectStyles($dom, $styles) {
```
jdemaris, line 57: I'd love to see type hints here for $dom to make it clearer, or at least PHPDoc comments

Empire/PrefillAdsInjector.php
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Empire;
+
+
+class PrefillAdsInjector {
+    private $adsConfig;
+    private $prefillConfig;
+    private $targeting;
+
+    public function __construct( $adsConfig, $prefillConfig, $targeting ) {
```
jdemaris, line 11: If we aren't making a class out of the AdsConfig, then we should at least document the expected keys / keys we depend on and what should be in them.

jdemaris (changes requested):
A few minor adjustments, but the biggest one is support for the non-standard post types that we allow in the configuration.

Empire/AdsInjector.php
```diff
@@ -0,0 +1,114 @@
...
+        return $importFragment;
+    }
+
+    public function checkAdsBlocked($adRules, $targeting) {
+        foreach ($adRules as $rule) {
+            if (!$rule['enabled']) {
+                continue;
+            }
+
+            $url = $targeting['url'];
```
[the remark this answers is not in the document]
sergunich: I made it variable mostly for readability

sergunich:
Thanks for the review @jdemaris! A lot of good catches and ideas!

Empire/PageInjection.php
ZLevine: Classic PHP 😆

PR #21, opened 2021-11-02
EP-874: Enable GraphQL Params

jdemaris:
Expose our server-side configs into GraphQL, if GraphQL is installed and enabled.

This is the first step toward native support for NextJS integrations.


Empire/GraphQL.php
```diff
@@ -0,0 +1,157 @@
...
+                    'connatixPlayspaceId' => $this->empire->useConnatix() ?
+                        $this->empire->getConnatixPlayspaceId() : null,
+                    'oneTrustEnabled' => $this->empire->useCmpOneTrust(),
+                    'oneTrustSiteId' => $this->empire->getOneTrustId(),
+                    'preloadConfigEnabled' => $this->empire->useInjectedAdsConfig(),
+                    'preloadConfigRules' => $this->empire->getAdsConfig()->adRules ?
+                        json_encode($this->empire->getAdsConfig()->adRules) : '[]',
+                    'preloadContainersEnabled' => $this->empire->useAdsSlotsPrefill(),
+                    'preloadContainersConfig' => $this->empire->getAdsConfig()->forPlacement ?
+                        json_encode($this->empire->getAdsConfig()->forPlacement) : '[]',
```
sergunich, line 136: Probably it will be settled with time how and what options are exposed, but it looks like several configs are mixed up here.

Currently, we have three separate configs:
1) AdsConfig - contains raw AdConfig that is inlined on the page, placements, and adRules that are used by the plugin internally
2) AmpConfig - contains HTML blobls of AMP components for placements
3) PrefillConfig - contains HTML blobs and styles for ad-conatiners

And I think they should be exposed independently.

sergunich (approved):

Empire/GraphQL.php
jdemaris: That makes a lot of sense! i'll adjust

PR #22, opened 2021-11-02
Add code formatting to build process

jdemaris:


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,29 @@
...
+          git clone -b master --depth 1 https://github.com/WordPress/WordPress-Coding-Standards.git $SNIFFS_DIR && \
+          $PHPCS_DIR/bin/phpcs --config-set installed_paths $SNIFFS_DIR
+
+      - uses: actions/checkout@v2
+
+      - name: Run PHPCS - Wordpress Plugin
+        run: $PHPCS_DIR/bin/phpcs --standard=WordPress empire.php
+
+      - name: Run PHPCS - PHP SDK
+        run: $PHPCS_DIR/bin/phpcs --standard=PSR12 Empire
```
sergunich, line 29: Probably this should be 
```
$PHPCS_DIR/bin/phpcs --standard=./src/phpcs.xml
```

sergunich (approved):
Awesome! This is great to have a linter for the plugin!

Two suggestions:
- Use composer for running linter locally. Similar to how it's done here https://github.com/empireio/solutions-wordpress/blob/master/src/composer.json#L81-L96
- Properly apply custom "standard" phpcs.xml, because currently looks like original Wordpress Styleguide is used. I would prefer not to use it as-is

.github/workflows/lint.yaml
jdemaris: Agreed - will tackle in #21

PR #31, opened 2021-11-26
EP-837 Add Campaigns support

rkashapov:
This PR adds to the WordPress plugin an ability to assign campaign assets to content.
https://recurrentventures.atlassian.net/browse/EP-837
Release when https://github.com/orgnc/platform/pull/652 is released


Empire/Campaigns.php
```diff
@@ -0,0 +1,38 @@
...
+
+        $assets = $this->empire->loadCampaignsAssets();
+        foreach ($assets as $asset) {
+            $field['choices'][$asset['guid']] = (
+                "#{$asset['externalId']} {$asset['name']} [{$asset['campaign']['name']}]"
+            );
+        }
+
+        return $field;
+    }
```
sergunich, line 37: It should be possible to use Organic WP Plugin with any WP installation not only with SWP and we should not rely on assumptions about how data is stored/saved in WP installation. With this approach we kind of force the client to use ACF fields in their WP.

So, in the case of SWP, this part of the code should go to the SWP codebase *and the plugin should only provide a method to pull assets.* That's what @jdemaris originally suggested [in the EP-837](https://recurrentventures.atlassian.net/browse/EP-837?focusedCommentId=21183):
> add a function (maybe empire_get_all_campaigns) that can fetch a list of campaigns from the Campaigns API (restricted to the current org, of course)

But instead of a global function, I would suggest [to use something like](https://github.com/orgnc/wordpress-plugin/blob/master/Empire/Empire.php#L233):
```php
Empire\Empire::getInstance()->getAllCampaigns()
```
What do you think @rrdial?

rkashapov: Thanks for the review, @sergunich 👍 

> But instead of a global function, I would suggest to use something like:
```php
Empire\Empire::getInstance()->getAllCampaigns()
```

I think global function provides better incapsulation.

Empire/CampaignAsset.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace Empire;
+
+class Campaign {
```
rrdial, line 5: The PHP autoloader needs this class to be in its own file.

rkashapov: Thank you, Ryan! Yesterday we spend a couple of hours investigating why this file isn't loaded. 🙂

Mwindo (approved):
Do we anticipate much else going in public.php? If so, it might be worthwhile making this a new directory so we can split functionality out, e.g. public/Campaigns.php vs. public.php. Otherwise, as far as I can tell (I've taken some time to learn the basics, but I'm no PHP expert!), this looks good.

empire.php
```diff
@@ -10,6 +10,7 @@
  */
 require __DIR__ . '/vendor/autoload.php';
 require __DIR__ . '/Empire/helpers.php';
+require __DIR__ . '/Empire/public.php';
```
rrdial, line 13: Not really scoped to this PR, but these files should be included in the [autoload](https://getcomposer.org/doc/04-schema.md#files) configuration.

rkashapov: Thank you, Ryan. Done

sergunich (approved):
Looks good! Thanks!

rkashapov:
Hi @Mwindo ,
> Do we anticipate much else going in public.php? If so, it might be worthwhile making this a new directory so we can split functionality out, e.g. public/Campaigns.php vs. public.php. 

I think for now we can start with one file. When the size of public.php is too big we'll split it. Before that, I'd prefer not to complicate the structure of files.

Thanks

PR #34, opened 2021-12-27
Fix assets query in Organic WordPress plugin

PetrMitsel:


sergunich (approved):
Thanks, @PetrMitsel! While fix looks minuscule, I'm sure that it was quite a challenge to make all these projects/repos work together :+1:

PR #48, opened 2022-02-09
Rename to Organic

jdemaris:
Rename everything feasible to Organic instead of Empire.

This includes fallback checks to read `empire::` from settings in the DB so we don't break during the transition and continued use of the `empire-wordpress-plugin` S3 bucket.

This also strips out the last vestiges of TrackADM


organic.php
```diff
@@ -0,0 +1,26 @@
...
+    /* @deprecated */
+    $environment = getenv( 'EMPIRE_ENVIRONMENT' );
+
+    if ( ! $environment ) {
+        $environment = 'PRODUCTION';
+    }
+}
+
+$organic = new Organic( $environment );
+$organic->init();
```
sergunich, line 26: Arguments for `init` missed here:
```php
$empire->init( getenv( 'EMPIRE_API_URL' ), getenv( 'EMPIRE_CDN_URL' ) );
```

Organic/Organic.php
```diff
@@ -1,124 +1,24 @@
...
-
-
-class LogLevel {
-    const WARNING   = 'warning';
-    const INFO      = 'info';
-    const DEBUG     = 'debug';
-}
+const CAMPAIGN_ASSET_META_KEY = 'organic_campaign_asset_guid';
+const GAM_ID_META_KEY = 'organic_gam_id';
+const SYNC_META_KEY = 'organic_sync';
```
sergunich, line 15: I think these post meta keys should be handled in a compatible manner, similar to `$this->getOption`, and be taken into account in a couple of SQL queries later in the code.

Even better would be to add some one-time script that can be run manually to migrate options and postmetas to the new keys

sergunich (approved):
I left several comments in the code, but overall - this is awesome! Much needed cleanup and re-branding of the plugin :slightly_smiling_face: 

Besides that, I think we should bump the MAJOR part of the version, because a lot of things changed, and mention in the release or README that usage of public API (`Organic/public.php`), filters, and ENV VARS should be updated/adjusted in the user's codebase.

By the way, I think that's a good opportunity to update/rename the plugin name and repo. Maybe to something like `organic-wp-plugin`

Organic/Api.php
```diff
@@ -1,38 +1,14 @@
 <?php
```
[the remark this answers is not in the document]
jdemaris: Good catch - removing

organic.php
jdemaris: bringing them back - they aren't used in production so I removed them, but I see they have other use cases.

PR #51, opened 2022-02-11
SWP: New Blank Page

regular-alves:
This change will not affect the production environment since the WordPress `apply_filters` function returns the 2nd parameter value when there isn't any place calling the filter. 
I'm implementing a new page template on the Empire theme where should not inject ads. So, the theme will use this filter to overlay the plugin ads injection settings on this template.

### DIFF
--- a/Organic/PageInjection.php
+++ b/Organic/PageInjection.php
@@ -89,7 +89,7 @@ class PageInjection {
                             return $content;
                         }
 
-                        if ( ! apply_filters( 'organic::eligible-for-ads', $this->organic->eligibleForAds( $content ) ) ) {
+                        if ( ! apply_filters( 'organic_eligible_for_ads', $this->organic->eligibleForAds( $content ) ) ) {
                             return $content;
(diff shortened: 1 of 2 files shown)


jdemaris (approved):
Nice work!

sergunich:
Hi @regular-alves, that's a clean solution! 
But for which task are these changes? I want to better understand the context of the task and why we need this.

regular-alves:
@sergunich, I created a new page template in Empire theme in which, even if I check the option that I want to inject ads on pages, it should not display ads.
You can check more information here https://recurrentventures.atlassian.net/jira/software/c/projects/SWP/boards/40?modal=detail&selectedIssue=SWP-1721

sergunich:
@regular-alves, okay, makes sense :+1: 
But looks like you forgot to apply_filter in case of AMP
https://github.com/orgnc/wordpress-plugin/blob/master/Organic/PageInjection.php#L60

PR #60, opened 2022-05-12
Add Affiliate feature

rkashapov:


sergunich (approved):
Thank Rustam! That's great work!
Today I learned how to add custom Gutenberg blocks 🙂

Organic/Affiliate.php
```diff
@@ -0,0 +1,36 @@
...
+    public function __construct( Organic $organic ) {
+        $this->organic = $organic;
+        add_action( 'init', [ $this, 'register_gutenberg_block' ] );
+        add_action( 'wp_enqueue_scripts', [ $this, 'register_scripts' ] );
+        add_action( 'admin_enqueue_scripts', [ $this, 'register_scripts' ] );
+    }
+
+    public function register_scripts() {
+        $siteId = $this->organic->getSiteId();
+        $sdk_url = getenv( 'ORGANIC_SDK_URL', 'https://organiccdn.io/assets/sdk/sdkv2.js' );
```
Mwindo, line 20: I'm probably just missing something, but I don't think a default value is allowed as the second argument to [`getenv`](https://www.php.net/manual/en/function.getenv.php)?

Mwindo (approved):
Great work, @rkashapov!

Organic/Affiliate.php
rkashapov: You're right! Nice catch 👍

g1r0 (approved):
Nicely done, Rustam! 💪🏻

PR #61, opened 2022-05-17
EP-1937: Adding functionality to initialize AdThrive code injection, if in the correct bucket. 

NitishMudgal:
**Change Request:**
1. Adding functionality to AB test against AdThrive plugin. 
2. Based on the bucketing values, the plugin should initialize AdThrive script load or load Organic Ad scripts.
3. AB testing key should contain `%organic_adthrive%` for this to run else, Organic scripts will load as usual.
4. Dynamically loading `gpt.js` to prevent multiple google tag script load while testing with other Ad Wordpress plugins.

**Dependencies:**
1. AdThrive Plugin needs to be updated. IIFE that loads ad scripts in file `adthrive-ads/components/ads/partials/ads.php` need to declared as a function. example:
```
<script>
window.loadAdThrive = (function(w, d) {
	w.adthrive = w.adthrive || {};
	w.adthrive.cmd = w.adthrive.cmd || [];
	w.adthrive.plugin = 'adthrive-ads-<?php echo esc_js( ADTHRIVE_ADS_VERSION ); ?>';
	w.adthrive.host = 'ads.adthrive.com';
	w.adthrive.integration = 'plugin';

	var commitParam = (w.adthriveCLS && w.adthriveCLS.bucket !== 'prod' && w.adthriveCLS.branch) ? '&commit=' + w.adthriveCLS.branch : '';

	var s = d.createElement('script');
	s.async = true;
	s.referrerpolicy='no-referrer-when-downgrade';
	s.src = 'https://' + w.adthrive.host + '/sites/<?php esc_attr_e( $data['site_id'] ); ?>/ads.min.js?referrer=' + w.encodeURIComponent(w.location.href) + commitParam + '&cb=' + (Math.floor(Math.random() * 100) + 1) + '<?php echo 'true' === $data['plugin_debug'] ? '&debug=true' : ''; ?>';
	var n = d.getElementsByTagName('script')[0];
	n.parentNode.insertBefore(s, n);
});
</script>
```

**Testing Flow:**
- [ ] Load Organic plugin on one of the solutions sites.
- [ ] check to see the ads load fine along with google publisher tags. You can check it by running `googletag.openConsole()`  in the console.
- [ ] Add AdThrive Plugin https://wordpress.org/plugins/adthrive-ads/ , update the `Site Id` value to `591b55faa29bbe3a6f50b09c`
- [ ] make updates to AdThrive as described in Dependency 1
- [ ] set `Key-Value for Split Test:` in Organic Plugin settings and use a string which contains `organic_adthrive` substring. 
- [ ] reload the page, either of the js scripts should load based on testing bucket. For AdThrive, `ads.min.js` should load
- [ ] confirm if the correct Google Publisher Tags are loaded using `googletag.openConsole()`

### DIFF
--- a/Organic/PageInjection.php
+++ b/Organic/PageInjection.php
@@ -2,6 +2,7 @@
 
 
 namespace Organic;
+
 /**
  * Handles adding data into the various pages on the website based on the selected
  * configuration.
@@ -315,11 +316,11 @@ class PageInjection {
                     <?php } ?>
                         }
 
-                <?php if ( $this->organic->useAdsSlotsPrefill() ) { ?>
+                    <?php if ( $this->organic->useAdsSlotsPrefill() ) { ?>
                         loadAds();
-                <?php } else { ?>
+                    <?php } else { ?>
                         setTimeout(loadAds, loadDelay);
-                <?php } ?>
+                    <?php } ?>
                     })();
                 }
             </script>
--- a/Organic/PageInjection.php
+++ b/Organic/PageInjection.php
@@ -2,7 +2,6 @@
 
 
 namespace Organic;
-
 /**
  * Handles adding data into the various pages on the website based on the selected
  * configuration.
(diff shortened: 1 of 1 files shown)


jdemaris (approved):
This looks good! I think we will want to make a more sophisticated control system for the tests later, but this should cover initial needs and establish a baseline framework.

PR #66, opened 2022-06-06
EP-1610: connatix on amp

glebofff:


sergunich (commented):
Looks good! And thanks for code cleanup/refactoring 👍

PR #64, opened 2022-05-24
Affiliate features

rkashapov:

PR #68, opened 2022-06-06
EP-2213: Affiliate AMP Support

Mwindo:
Based off of https://github.com/orgnc/wordpress-plugin/pull/64. Allows us to render a product card in an amp-iframe.

Example:
Screen Shot 2022-06-08 at 6 12 49 PM


affiliate/src/productCard/Edit.jsx
```diff
@@ -179,6 +186,12 @@ const Edit = ({ attributes, setAttributes, productSearchPageUrl }) => {
                 label="Display Description"
                 onChange={setDisplayDescription}
               />
+              <CheckboxControl
+                checked={attributes.isAmp}
+                help="Whether or not the product card should support AMP"
```
Mwindo, line 191: I should maybe add a note like "(AMP plugin must be enabled)"

affiliate/src/productCard/ProductCard.jsx
```diff
@@ -18,26 +18,44 @@ const ProductCard = forwardRef(({
...
+    return (
+      <div
+        ref={ref}
+        data-organic-affiliate-integration="product-card"
+        data-organic-affiliate-integration-options={options}
+        data-organic-affiliate-product-guid={productGuid}
+      />
+    );
+  }
+  const src = `${publicDomain}/integrations/affiliate/product-card?guid=${productGuid}&${options}`;
```
rkashapov, line 49: Shouldn't we translate div into amp-iframe in the Affiliate SDK instead?

Mwindo: Can we? I was under the impression that the Affiliate SDK wouldn't be able to run on AMP pages (at least yet) since arbitrary JavaScript isn't allowed, and it doesn't look like our SDK code is embedded using `amp-script`. If we can, then I am all for doing it via the SDK.

rkashapov: > I was under the impression that the Affiliate SDK wouldn't be able to run on AMP pages

Probably you're right. I don't know what AMP is. We translate `div` into `iframe` by Affiliate SDK and thought that doing so would be consistent. If it isn't possible technically, then it's ok

affiliate/src/productCard/ProductCard.jsx
```diff
@@ -18,26 +18,44 @@ const ProductCard = forwardRef(({
...
+      displayDescription,
+      displayImage,
+      cardRadius,
+      cardShadow,
+      textColor: convertColorValue(textColor),
+      linkColor: convertColorValue(linkColor),
+      backgroundColor: convertColorValue(backgroundColor),
+    },
+  );
+  if (!isAmp) {
```
rkashapov, line 39:
```suggestion
  if (isAmp) {
    return <amp-iframe
  }
```

affiliate/src/productCard/ProductCard.jsx
rkashapov: I discussed how AMP works in the wordpress plugin with @sergunich . The same page could be rendered for AMP and not. For instance, for desktop browsers a page is rendered in the regular mode and in AMP mode for mobile devices. We should determine how to render the product card in runtime. We can do this on backend side.

In the PageInjector class we have amp-dependent logic
https://github.com/orgnc/wordpress-plugin/blob/master/Organic/PageInjection.php#L28-L51

What do you think if we'll insert an `amp-iframe` inside the `div[data-organic-affiliate-integration="product-card"]` elements there? I think `injectConnatix` is a good example of how it could be done.
https://github.com/orgnc/wordpress-plugin/blob/50ce425e6e0a30461371e9bdf4e0c578f9fc4beb/Organic/AmpAdsInjector.php#L65-L93

Mwindo: Ah, ok, this makes sense. I will switch to this. I agree that this is better.

Organic/AmpAdsInjector.php
```diff
@@ -3,8 +3,22 @@
 namespace Organic;
```
Mwindo, line 3: These changes just reflect the current master branch.

Organic/AmpAffiliateInjector.php
```diff
@@ -0,0 +1,55 @@
...
+            <amp-iframe
+                height="540px"
+                frameborder="0"
+                sandbox="allow-scripts allow-same-origin allow-popups allow-popups-to-escape-sandbox"
+                src="$url"
+            >
+                <p placeholder>Loading iframe content</p>
+            </amp-iframe>
+        HTML;
+        $amp_iframe = $this->dom->createElement( $amp_iframe_code );
```
Mwindo, line 51: This is nonsense

Organic/AmpAdsInjector.php
rkashapov: Don't include changes from the master branch in your commits please. We'll have problems with merging them.
Use rebase instead. I've rebased the `feature/affiliate-link` branch, please update yours as well. Thank you

Organic/AmpAffiliateInjector.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace Organic;
+
+use AMP_Base_Sanitizer;
```
Mwindo, line 5: @sergunich This was the issue earlier--AMP_Base_Sanitizer wasn't included.

sergunich: Huh... yeah, this namespace resolution thing in PHP is a bit weird

Organic/AmpAffiliateInjector.php
```diff
@@ -0,0 +1,61 @@
...
+use DOMDocument;
+use DOMXPath;
+
+class AmpAffiliateInjector extends \AMP_Base_Sanitizer {
+    private $organic;
+
+    public function sanitize() {
+        try {
+            $this->organic = Organic::getInstance();
+            $this->site_organic_domain = $this->organic->getSiteOrganicDomain();
```
rkashapov, line 15: The organic object doesn't have `getSiteOrganicDomain` method. I removed it since the SDK has inlined configuration.

Organic/AmpAffiliateInjector.php
```diff
@@ -0,0 +1,61 @@
...
+    public function handle() {
+        if ( empty( $this->site_organic_domain ) ) {
+            return; // can't insert amp-iframes without a public domain
+        }
+        $this->handle_product_cards();
+    }
+
+    public function handle_product_cards() {
+        $xpath = new DOMXPath( $this->dom );
+        $product_card_divs = $xpath->query( "//div[@data-organic-affiliate-integration='product-card']" );
```
rkashapov, line 31: Shouldn't we check/add `data-organic-affiliate-processed` attribute?

Organic/AmpAffiliateInjector.php
Mwindo: I think this means I'll need to do create a new platform endpoint to execute a GraphQL query for the site domain. Does that sound right? I want to make sure you don't have an easier approach.

rkashapov: We have an enpoint that is used to inline configs, you can use that one:
https://github.com/orgnc/platform/blob/9d1c8b0143b77cf2bf90cff6363bb62f24cbe380/backend/sdk/api/rest.py#L13

Mwindo: Ah, awesome. Thank you!

Mwindo:
Opened separate branch https://github.com/orgnc/wordpress-plugin/pull/86

PR #80, opened 2022-06-10
EP-1610: amp-connatix-player custom params and macros

glebofff:


Organic/AmpAdsInjector.php
```diff
@@ -69,13 +69,44 @@ private function injectConnatix() {
...
+            ),
+            ENT_QUOTES,
+            'UTF-8'
+        );
+
         $psid = $this->organic->getConnatixPlayspaceId();
         $player = "
             <amp-connatix-player
                 data-player-id=\"ps_$psid\"
+                data-param-playback-mode=\"1\"
```
sergunich, line 102: Do we really need to explicitly set playback mode?
It's not set in regular Connatix injection code, and I believe in this case playback mode is controlled by configuration in Connatix Portal

sergunich (approved):
Looks good! Only one minor question

Organic/AmpAdsInjector.php
glebofff: Ok

PR #90, opened 2022-07-03
SWP-2781 Content sync: prioritize modification date from metadata

pasenor:
Mimicking [this platform logic](https://github.com/orgnc/solutions-wordpress/blob/2767f65c4bb097e9a862d114708bd9da1a299bb2/src/web/app/mu-plugins/swp-core/EmpireSWP/CustomPostModifiedTimestamp.php#L407)

### DIFF
--- a/Organic/Organic.php
+++ b/Organic/Organic.php
@@ -741,7 +741,11 @@ class Organic {
         $content = \apply_filters( 'organic_post_content', $post->post_content, $post->ID );
         $is_published = \apply_filters( 'organic_post_is_published', $post->post_is_published, $post->ID );
         $published_date = \apply_filters( 'organic_post_publish_date', $post->post_date, $post->ID );
-        $modified_date = \apply_filters( 'organic_post_modified_date', $post->post_modified, $post->ID );
+        $modified_date = \apply_filters(
+            'organic_post_modified_date',
+            \apply_filters( 'get_display_datetime', $post->ID ),
+            $post->ID
+        );
         $campaign_asset_guid = null;
         if ( $this->isCampaignsAppEnabled() ) {
             $campaign_asset_guid = get_post_meta( $post->ID, CAMPAIGN_ASSET_META_KEY, true );


Organic/Organic.php
```diff
@@ -740,7 +742,13 @@ public function syncPost( $post ) {
         $sponsorship = \apply_filters( 'organic_post_sponsorship', $post->post_sponsorship, $post->ID );
         $content = \apply_filters( 'organic_post_content', $post->post_content, $post->ID );
         $is_published = \apply_filters( 'organic_post_is_published', $post->post_is_published, $post->ID );
-        $published_date = \apply_filters( 'organic_post_publish_date', $post->post_date, $post->ID );
+
+        $post_meta_modified_date =  get_post_meta( $post->ID, STORY_TIMESTAMP_META_KEY, true );
```
sergunich, line 746: Organic WP plugin could be used not only by SWP but by any client with Wordpress installation. So, we shouldn't rely or depend on SWP internal logic.

A similar problem was discussed here: https://github.com/orgnc/wordpress-plugin/pull/31#discussion_r759560774

Instead of updating the plugin code, the correct `modified_date` should be passed from the SWP side through WP filters. Something like this: https://github.com/orgnc/solutions-wordpress/blob/master/src/web/app/themes/empire/app/filters.php#L2003

sergunich (changes requested):

Organic/Organic.php
```diff
@@ -741,7 +741,11 @@ public function syncPost( $post ) {
         $content = \apply_filters( 'organic_post_content', $post->post_content, $post->ID );
         $is_published = \apply_filters( 'organic_post_is_published', $post->post_is_published, $post->ID );
         $published_date = \apply_filters( 'organic_post_publish_date', $post->post_date, $post->ID );
-        $modified_date = \apply_filters( 'organic_post_modified_date', $post->post_modified, $post->ID );
+        $modified_date = \apply_filters(
+            'organic_post_modified_date',
+            \apply_filters( 'get_display_datetime', $post->ID ),
```
sergunich, line 746: Not sure why we need another filter here? I think using the `organic_post_modified_date` should be enough
Similar to this: https://github.com/orgnc/api.thedrive.com/blob/04cf6b75fd1ece0e78f54c17a062d6f4fb3facb3/src/wp-content/plugins/empire-integration/empire-integration.php

That's an outdated repo with an old plugin (which was named `empire` at that time) but should give a general idea

pasenor:
Ticket implemented in SWP without changes in the plugin.

PR #95, opened 2022-07-21
Fix WordPress RSS Feed

enescala:


iceone (approved):

Organic/PrefillAdsInjector.php
```diff
@@ -21,6 +21,9 @@ public function __construct( AdsConfig $adsConfig, PrefillConfig $prefillConfig,
     }
 
     public function prefill( $content ) {
+        if ( is_feed() ) {
```
sergunich, line 24: It would be better to put this check here
https://github.com/orgnc/wordpress-plugin/blob/master/Organic/Organic.php#L429

PR #99, opened 2022-08-01
SWP-2832 button in plugin settings to sync categories

pasenor:


Organic/ContentSyncCommand.php
```diff
@@ -77,8 +77,6 @@ public function __invoke( $args, $opts ) {
             return;
         }
 
-        $this->organic->syncCategories();
```
sergunich, line 80: Why not to sync catergories periodically?

pasenor: They were failing and blocking all content sync. It should be fixed with this PR, but just to be sure I'd like to observe how it works, then after a while return this line.

pasenor: On second thought it doesn't seem a good idea, put it back. Still worth keeping an eye on the content sync for some days after the release.

sergunich (approved):
Looks good!

PR #101, opened 2022-08-04
EP-2824: connatix sdk settings

glebofff:


Organic/ConnatixConfig.php
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * @var string
+     */
+    private $playspaceId;
+
+    public function __construct( Organic $organic ) {
+
+        $rawAdsConfig = $organic->getOption( 'organic::ad_settings', [] );
+        $raw = isset( $rawAdsConfig['adSettings']['connatix'] )
```
sergunich, line 31: As I mentioned [in the review for other PR](https://github.com/orgnc/platform/pull/2005#pullrequestreview-1063106753) we shouldn't use SDK-level config for Connatix.

Instead, I think we should:
1) Use Placement-level setting by default (as you are already doing)
2) Use Plugin-level setting only for backward compatibility if there was no Placement for Connatix in the Ad config

glebofff: Ok, makes sense, i'll close platform PR.

Organic/AmpAdsInjector.php
```diff
@@ -43,12 +62,29 @@ public function handle() {
...
 
-            // certain placement is blocked
-            if ( $rule && in_array( $key, $blockedKeys ) ) {
+            // if placement's ad_type is set to 'outstream_video':
+            // inject connatix player instead of generic amp template
+            if ( $placement['adType'] == AD_TYPE::OUTSTREAM_VIDEO ) {
+
+                // get playspaceId from placement's 'connatixId'
+                $connatixId = trim( $placement['connatixId'] ) ?: '';
+                $psid = is_valid_uuid( $connatixId ) ? $connatixId : '';
```
sergunich, line 80: If `$psid` ends up being an empty string probably we should skip injections of Connatix, because it will not work either way

sergunich (approved):
Looks good!

Organic/AmpAdsInjector.php
glebofff: [Yep, it's already done.](https://github.com/orgnc/wordpress-plugin/blob/c3ab7e40ce1c10c476fa7afe3d852941f171ebec/Organic/AmpAdsInjector.php#L113)

PR #103, opened 2022-08-09
Remove ads.txt sync functionality; for requests to /ads.txt issue a r…

omonbek22:
…edirect to Site GUID's specific ads.txt


Organic/Organic.php
```diff
@@ -287,12 +287,10 @@ public function init( $apiUrl = null, $cdnUrl = null ) {
         /* Load up our sub-page configs */
         new AdminSettings( $this );
         new CCPAPage( $this );
-        new AdsTxt( $this );
```
sergunich, line 290: Have you tested these changes?
The `AdsTxt` ads custom action to  WP `init` filter during the initialization:
https://github.com/orgnc/wordpress-plugin/blob/d878e613952990f212f426124be6e3a7a9369e68/Organic/AdsTxt.php#L22

So.. without initialization of `AdsTxt` all of this won't work I suppose

Organic/AdsTxt.php
```diff
@@ -22,24 +23,18 @@ public function __construct( Organic $organic ) {
     }
 
     public function get() {
```
sergunich, line 25: I think we should rename the method considering that it just returns URL where to redirect

Organic/AdsTxt.php
```diff
@@ -22,24 +23,18 @@ public function __construct( Organic $organic ) {
...
     public function show() {
         if ( isset( $_SERVER ) && $_SERVER['REQUEST_URI'] === '/ads.txt' ) {
             $enabled = $this->organic->getOption( 'organic::enabled' );
 
             if ( $enabled ) {
-                $adsTxt = $this->organic->getOption( 'organic::ads_txt' );
-                header( 'content-type: text/plain; charset=UTF-8' );
-                header( 'cache-control: public, max-age=86400' );
-                echo $adsTxt;
+                header( 'Location: ' . $this->get() );
```
sergunich, line 34: Oh! And probably we should add a comment that only one redirect is allowed for `/ads.txt`
Per Ads.txt specification:
> Only a single HTTP redirect to a destination outside the original root domain is allowed to facilitate one-hop delegation of authority to a third party's web server domain.

It's not an issue in this case but comment could help prevent problems in future when someone else will work on this

Organic/AdsTxt.php
omonbek22: Agreed - will rename.

Organic/AdsTxt.php
omonbek22: I'll add a comment.

Organic/Organic.php
sergunich: Oh! I see, yeah. Didn't saw that 👍

Organic/AdsTxt.php
```diff
@@ -21,25 +22,26 @@ public function __construct( Organic $organic ) {
         add_action( 'init', array( $this, 'show' ) );
     }
 
-    public function get() {
-         return $this->organic->getOption( 'organic::ads_txt' );
+    public function getAdsTxtUrl() {
```
sergunich, line 25: Looks like there is one more place where `ads.txt` was used
https://github.com/orgnc/wordpress-plugin/blob/feature%2FEP-822%2Fplugin-ads-redirect/Organic/GraphQL.php#L144

This GQL API is used by `organic-web`
https://github.com/orgnc/organic-web/blob/master/src/lib/wordpress/api/fragments/organicConfigFragment.ts

Looks like this field is actually not used there, but you should check with Content Team to be sure and ask them about how they handle the `ads.txt`

glebofff (approved):
Overall, looks good and - i checked - it works. Good job. :-)

Organic/AdminSettings.php
```diff
@@ -320,7 +343,19 @@ public function showSettings() {
                 <input type="hidden" name="organic_sync_ads_txt" id="organic_sync_ads_txt" value="true" />
                 <p><input type="submit" value="Sync ads.txt"/></p>
             </form>
-
+            </div>
+            <script>
+                var hideShowSyncAdsTxt = function() {
```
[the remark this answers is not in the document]
omonbek22: updated

Organic/AdsTxtSyncCommand.php
```diff
@@ -12,6 +12,10 @@ class AdsTxtSyncCommand {
 
     public function __construct( Organic $organic ) {
         $this->organic = $organic;
+        $ads_txt_redirect = $this->organic->getOption( 'organic::ads_txt_redirect_enabled' );
```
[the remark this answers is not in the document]
omonbek22: updated

Organic/AdminSettings.php
```diff
@@ -94,7 +94,7 @@ protected function syncSettings() {
         if ( $_SERVER['REQUEST_METHOD'] != 'POST' ) {
             return;
         }
-        if ( ! isset($_POST['organic_update'] ) ) {
+        if ( ! isset( $_POST['organic_update'] ) ) {
```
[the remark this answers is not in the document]
omonbek22: squashed

PR #111, opened 2022-09-23
EP-3358: Removing redundant ad setup call, fixing if statement condition

NitishMudgal:
**Change Request:**
We need to control how much traffic Organic Ads are loaded on for initial client onboarding. One instance is when the client are still working with other Ad Stack and we want to load our ads.txt entries without rendering any ads.
 
1. Organic Ad traffic control with 0% traffic was breaking as the condition 
`$this->organic->getOrganicPixelTestPercent() != ''` was comparing empty string to integer. This would always result in false and not create `BVTests` object.
Screen Shot 2022-09-24 at 1 29 07 AM

2. Removed redundant AB testing code which instantiated other AdStack.

### DIFF
--- a/Organic/PageInjection.php
+++ b/Organic/PageInjection.php
@@ -415,8 +415,7 @@ class PageInjection {
                 <?php
                 if (
                         $this->organic->getOrganicPixelTestValue() &&
-                        $this->organic->getOrganicPixelTestPercent() != '' &&
-                        $this->organic->getOrganicPixelTestPercent() != null
+                        $this->organic->getOrganicPixelTestPercent() !== null
                 ) {
                     ?>
                 window.organicTestKey = "<?php echo $this->organic->getOrganicPixelTestValue(); ?>";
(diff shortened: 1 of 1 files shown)


Organic/PageInjection.php
```diff
@@ -415,8 +415,7 @@ function t() {
                 <?php
                 if (
                         $this->organic->getOrganicPixelTestValue() &&
-                        $this->organic->getOrganicPixelTestPercent() != '' &&
-                        $this->organic->getOrganicPixelTestPercent() != null
+                        $this->organic->getOrganicPixelTestPercent() !== null
```
jdemaris, line 418: Can we at least check if it's a number? It will likely mess up the `enabled: <value>` Javascript if not

jdemaris (approved):
Overall this looks fine. Ideally we have one extra safety check.

Organic/PageInjection.php
NitishMudgal: `organicPixelTestPercent` is defined using `intval` function which will return 0 for failure and int value in case of successful string->int conversion.
https://github.com/orgnc/wordpress-plugin/blob/4f02059389460bf33882a093135cc1948d0be920/Organic/Organic.php#L263

PR #121, opened 2022-10-28
EP-3665 Add banner text to product card plugin

amittman:
Adding the ability to add banner/award to the product card plugin. Also, needs this [PR](https://github.com/orgnc/wordpress-plugin/pull/121) to work properly with the current bannerText data attribute setup

https://user-images.githubusercontent.com/9379158/198701186-78fa5d5e-2fa4-4c74-8e00-dedee4364f8b.mov

### DIFF
--- a/affiliate/package.json
+++ b/affiliate/package.json
@@ -21,7 +21,9 @@
 		"eslint-plugin-react": "7.30.0"
 	},
 	"dependencies": {
+		"@wordpress/element": "^4.18.0",
 		"@wordpress/scripts": "23.0.0",
+		"lodash": "^4.17.21",
 		"prop-types": "15.8.1"
 	}
 }
--- a/affiliate/src/productCard/Edit.jsx
+++ b/affiliate/src/productCard/Edit.jsx
@@ -7,15 +7,18 @@ import {
   CardBody,
   CardDivider,
   CheckboxControl,
-  Toolbar,
   IconButton,
+  Toolbar,
+  TextControl,
 } from '@wordpress/components';
 import {
   createRef,
-  useState,
-  useEffect,
   useCallback,
+  useEffect,
+  useMemo,
+  useState,
 } from '@wordpress/element';
+import { debounce } from 'lodash';
 import PropTypes from 'prop-types';
 
 import ProductSearchModal from '../ProductSearchModal';
@@ -36,6 +39,7 @@ const Edit = ({ attributes, setAttributes, productSearchPageUrl }) => {
     attributes.productGuid,
     attributes.displayImage,
     attributes.displayDescription,
+    attributes.bannerText,
   ]);
   const [showModal, setShowModal] = useState(!attributes.productGuid);
   const hideModal = useCallback(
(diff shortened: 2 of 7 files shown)


affiliate/src/productCard/Edit.jsx
```diff
@@ -98,11 +108,18 @@ const Edit = ({ attributes, setAttributes, productSearchPageUrl }) => {
                 label="Display Description"
                 onChange={setDisplayDescription}
               />
+              <TextControl
+                help="Text for product card banner/award"
+                label="Banner Text"
```
amittman, line 113: @Mwindo I'm not sure how we want to market this feature. Is it "Banner" or "Award"? I just put whatever here, expecting we would want to change it. (I think I would also rather use "Banner" over "Banner Text" for the label).

Mwindo: Banner is more agnostic, so let's go with that. I like Banner Text slightly more since it is more descriptive, but Banner or Banner Text is fine with me!

affiliate/package.json
```diff
@@ -21,7 +21,9 @@
 		"eslint-plugin-react": "7.30.0"
 	},
 	"dependencies": {
+		"@wordpress/element": "^4.18.0",
```
amittman, line 24: I'm not sure why this wasn't included before... Maybe it just cuts down on the size of the plugin. I added this without thinking too much about it to fix a warning that I thought was causing the pr to fail (but I realized that warnings don't fail in this repo).

Starting to think that it might be best to remove this for now, since I haven't been able to dig up the reasoning for this missing...

affiliate/src/productCard/Edit.jsx
```diff
@@ -61,6 +67,20 @@ const Edit = ({ attributes, setAttributes, productSearchPageUrl }) => {
     (displayDescription) => setAttributes({ displayDescription }),
     [setAttributes],
   );
+
+  const [bannerTextValue, setBannerTextValue] = useState(attributes.bannerText);
```
amittman, line 71: I assumed that we didn't want to have a default text, except for whatever has been saved already. So, we wouldn't want something like:
```js
useState(attributes.bannerText || 'Best Overall')
```

affiliate/package.json
Mwindo: Might be worth having @sergunich chime in here as he has more WordPress plugin experience.

sergunich: > I'm not sure why this wasn't included before
That's because `@wordpress/scripts` has a `@wordpress/babel-preset-default` as dependency
https://github.com/orgnc/wordpress-plugin/blob/ffca2768419ca39ca304fdd5dcad70ddfe1a6b60/affiliate/package-lock.json#L3448

And `@wordpress/babel-preset-default` has a `@wordpress/element` as dependency
https://github.com/orgnc/wordpress-plugin/blob/ffca2768419ca39ca304fdd5dcad70ddfe1a6b60/affiliate/package-lock.json#L3265

So, it was installed implicitly, and yes - it's better to state it as a dependency explicitly

> Maybe it just cuts down on the size of the plugin
We don't care about the size of the plugin 🙂

sergunich (approved):

affiliate/package.json
amittman: Thanks for the explanation!

Mwindo: Thanks @sergunich :)

Mwindo (approved):
LGTM. Thanks, @amittman!

PR #128, opened 2022-11-30
EP-3250 Add Product Carousel 

amittman:


affiliate/webpack.config.js
```diff
@@ -0,0 +1,67 @@
...
+ * Resolve a block name to the path to it's main `index.js` entry-point.
+ * @param string name The name of the block.
+ * @returns string A normalized path to the block's entry-point file.
+ * */
+const resolveBlockEntry = (name) => resolveSource(name, 'index.js');
+
+config.entry = {
+  'productCard/index': resolveBlockEntry('productCard'),
+  'productCarousel/index': resolveBlockEntry('productCarousel'),
+};
```
amittman, line 42: I've been trying to get this directory to accept multiple blocks through a few different methods and this is the closest I have been able to get to that goal. At first, I was hoping it would be as simple as [this Github thread](https://github.com/WordPress/gutenberg/issues/40771#issuecomment-1127772731) makes it out to be rather simple but the build directory would be out of sorts whenever I tried it, so I started trying to change the Webpack build process. Which led me to [this stackexchange post](https://wordpress.stackexchange.com/questions/390282) but it is still not working as hoped.

Currently, the Product Carousel seems to be recognized by WP, can be placed on the page, and I see the modal working (haven't been able to fully test because it still uses the production version of the product carousel creation page and I only just now pushed the [PR](https://github.com/orgnc/platform/pull/2745)). However, the Product Card isn't showing up in WP. I also dislike using this Webpack config file, since it is pretty messy and would like to be able to use the method in that first Github link, but it has been a problem.

affiliate/src/components/IntegrationModal.jsx
```diff
@@ -0,0 +1,51 @@
+import './integrationModal.scss';
+
+import { Modal } from '@wordpress/components';
+import { useEffect } from '@wordpress/element';
+import PropTypes from 'prop-types';
+
+const IntegrationModal = ({
```
Mwindo, line 7: 💪

affiliate/package.json
```diff
@@ -21,6 +21,9 @@
 		"eslint-plugin-react": "7.30.0"
 	},
 	"dependencies": {
+		"@wordpress/block-editor": "^10.5.0",
+		"@wordpress/blocks": "^11.21.0",
+		"@wordpress/components": "^22.1.0",
```
amittman, line 26: Added these so they could join `@wordpress/element` in being explicitly defined.

affiliate/src/productCarousel/components/IntegrationModal.jsx
```diff
@@ -0,0 +1,51 @@
+import './integrationModal.scss';
+
+import { Modal } from '@wordpress/components';
+import { useEffect } from '@wordpress/element';
+import PropTypes from 'prop-types';
+
+const IntegrationModal = ({
```
amittman, line 7: Code re-use, but afaik, it's tricky to share code among these Gutenberg blocks, so it might be better for now to make changes to one "master" components directory and then copy/paste them into the other blocks until a better way is found 🤷

Mwindo: Does it not work pulling this component into a "shared" directory? It seems to work for me, but I could definitely be missing something. See comment I'm about to add below.

affiliate/webpack.config.js
Mwindo: I wonder what you think about this sort of refactor? https://github.com/orgnc/wordpress-plugin/pull/131. (Definitely needs cleaning up! For instance, if we did things this way, we'd need to fix the build-zip.sh.) I've gotten things to work with a single IntegrationModal, and changing package.json to run differently. Still getting the spacing weirdness though.

Mwindo: So instead of having one src directory, we have a blocks directory; inside, for each integration type, we have a directory with the block and src. Our package.json build script builds each of these separately.

Mwindo: In any case, this refactor gets rid of the config.

amittman: Man, I used that same sort of `wp-scripts build` line with the output-path but it didn't seem to work... Is it just because of the nested `blocks` directory? Or, I might just be at constant odds with this dev environment. Personally, I don't like the deeply nested structure and would like to keep the directory structure similar to this PR since these blocks aren't supposed to have a lot of moving parts so adding more nested directories ends up looking more confusing.

But, I'm good with that direction if the build script doesn't end up working with this PR's directory setup.

amittman: Also, I think this whitespace might be a problem with the dev environment. I'm not seeing what could have caused this and it appears even when I switch back to the master branch...

Mwindo: > Or, I might just be at constant odds with this dev environment.

I spent, like, another hour getting it working again, so it's not just you!

I _think_ the current directory structure should work. I think the following should do the trick:

- Be sure that there are two separate build directories (otherwise I believe one will overwrite the other), so, e.g., src/productCard/build and src/productCarousel/build. The build script (which I think we should rename to "affiliate-integrations") should run wp-scripts build <productCard> && wp-scripts build <productCarousel>.
- Make sure to repoint the register-scripts stuff in Affiliate.php to the new locations.

Mwindo: > Also, I think this whitespace might be a problem with the dev environment. I'm not seeing what could have caused this and it appears even when I switch back to the master branch...

Wondering this myself. I'll post to dev-environment.

affiliate/src/productCard/Edit.jsx
```diff
@@ -41,7 +41,7 @@ const Edit = ({ attributes, setAttributes, productSearchPageUrl }) => {
     attributes.displayDescription,
     attributes.bannerText,
   ]);
-  const [showModal, setShowModal] = useState(!attributes.productGuid);
+  const [showModal, setShowModal] = useState(false);
```
Mwindo, line 44: I like the behavior of the modal opening automatically upon first selecting the integration, in other words, without having to press the icon; however, I also see how this can be an issue since the same logic will open the modal whenever the editor visits a page that has an unfinished integration. Maybe we could add an attribute so that the modal pops up by default the first time the user is building the integration?

affiliate/src/productCarousel/Edit.jsx
```diff
@@ -0,0 +1,92 @@
...
+import PropTypes from 'prop-types';
+
+import ProductCarousel from './ProductCarousel';
+import ProductCarouselModal from './ProductCarouselModal';
+import { AttributesType } from './propTypes';
+
+const Edit = ({ attributes, setAttributes, productCarouselCreationURL }) => {
+  const productCarouselRef = createRef();
+
+  const [showModal, setShowModal] = useState(false);
```
Mwindo, line 23: Same comment as for the productCard Edit.jsx.

affiliate/src/productCarousel/Edit.jsx
```diff
@@ -0,0 +1,92 @@
...
+          Product Carousel
+        </CardHeader>
+        <CardBody>
+          {attributes.productCarouselSnippet ? (
+            <ProductCarousel
+              ref={productCarouselRef}
+              productCarouselSnippet={attributes.productCarouselSnippet}
+            />
+          ) : (
+            <h3>Product Carousel is not created</h3>
```
Mwindo, line 78: To ensure the user knows they need to do something (rather than this just being an error), maybe we say, "Product Carousel is unfinished."

Mwindo (commented):
This is great work, @amittman! I've left a couple more comments, and I'll do another pass once this is out of draft mode, but it's looking good.

affiliate/src/productCard/Edit.jsx
amittman: Yeah, I must have been visiting a page that had an unfinished one without realizing it. I reverted this.

build-zip.sh
```diff
@@ -12,7 +12,8 @@ cp composer.lock build/organic/
 cp -r vendor build/organic/
 (cd affiliate/ && npm install && npm run build)
 mkdir -p build/organic/affiliate/
-cp -r affiliate/build build/organic/affiliate/
+cp -r affiliate/blocks/productCard/build build/organic/affiliate/productCard
+cp -r affiliate/blocks/productCarousel/build build/organic/affiliate/productCarousel
 cd build
 zip -r organic-${BUILD_NUMBER}.zip organic
 cp organic-${BUILD_NUMBER}.zip organic.zip
```
amittman, line 19: I'm not sure how this build file should look. Winging it while I read up on it some more

Mwindo: This looks reasonable, but it's maybe something @sergunich has better insights into.

Mwindo (approved):
LGTM, but I definitely want @sergunich or someone to look over build-zip.sh. Thank you, @amittman, and nice work 🔥

albekov (approved):
LGTM 👍

build-zip.sh
amittman: Maybe @regular-alves could take a quick look at this build file. Amy mentioned that you might be able to take a quick look and confirm.

amittman: Thanks for taking a look!

Mwindo: I think this is actually pretty easy. We just need to copy the existing directory structure into build/organic. In other words, https://github.com/orgnc/wordpress-plugin contains `organic.php`, `Organic`, `affiliate`, etc., so `build/organic` should contain `organic.php`, `Organic`, `affiliate`, etc. `affiliate` itself contains `blocks/productCard` and `blocks/productCarousel` and `blocks/shared`, so these should be copied appropriately (so that, when this zip is uploaded into WordPress, the register_block_type path we've defined--based on blocks/...--works). So I think what we want is:

```
(cd affiliate/ && npm install && npm run build)
mkdir -p build/organic/affiliate/blocks/
cp -r affiliate/blocks/productCard/build build/organic/affiliate/blocks/productCard/
cp -r affiliate/blocks/productCarousel/build build/organic/affiliate/blocks/productCarousel/
cp -r affiliate/blocks/shared build/organic/affiliate/blocks/shared/
cd build
zip -r organic-${BUILD_NUMBER}.zip organic
cp organic-${BUILD_NUMBER}.zip organic.zip
```

That said, I'm still trying to find a good way to test.

amittman: Yeah, makes sense

Mwindo: My above was the right idea, but executed incorrectly. Here is what the code should be:

```
(cd affiliate/ && npm install && npm run build)
mkdir -p build/organic/affiliate/blocks/
cp -r affiliate/blocks/ build/organic/affiliate/
cd build
zip -r organic-${BUILD_NUMBER}.zip organic
cp organic-${BUILD_NUMBER}.zip organic.zip
```

This will recreate the directory structure correctly.

It turns out this is super trivial to test once you understand how it all works: delete wordpress-plugin/build (if applicable); run the build script in the container; replace the mu-plugins/wordpress-plugin directory with the newly built wordpress-plugin/build/organic directory. E.g., I moved wordpress-plugin to my Desktop temporarily, then moved its build/organic subdirectory into the mu-plugins directory and renamed it to wordpress-plugin. Everything should still work in the WP editor--after all, you've just replaced your working wordpress-plugin directory with what should be another working wordpress-plugin directory. I've confirmed that the above code works and I've confirmed that this testing procedure works.

Once this change is made, we can finally merge this PR. I'll update documentation because this is something that could easily cost others development time when it really shouldn't.

amittman: Awesome! Thanks for testing it

Mwindo: Confirming that the above works. I don't know why it wouldn't, but I've been staring at this so long that nothing makes sense, so good to double-check.

Mwindo: Also, @amittman, be sure to update the wordpress-plugin version in solutions-wordpress.

amittman: Yep, was waiting until this was ready to go

PR #132, opened 2022-12-10
Update README after affiliate changes

Mwindo:


amittman (approved):
LGTM, thanks again @Mwindo

PR #134, opened 2022-12-17
Bugfix: Fix Affiliate Link Insertion

Mwindo:
See https://organicventures.slack.com/archives/C01LLSNMKT2/p1671219880801499.

### DIFF
--- a/affiliate/blocks/productCard/insertLink/InsertAffiliateLink.jsx
+++ b/affiliate/blocks/productCard/insertLink/InsertAffiliateLink.jsx
@@ -35,9 +35,9 @@ const InsertAffiliateLink = ({
     () => setShowModal(true),
     [setShowModal],
   );
-  const onProductSelect = useCallback((productInfo) => {
+  // format of arguments is based on data received from postMessage event in orgnc/platform file IntegrationProductSearch.jsx
+  const onProductSelect = useCallback(({ product, offerUrl }) => {
     const { start, end } = value;
-    const { offerUrl, product } = productInfo;
     const format = {
       type: FORMAT_NAME,
       attributes: { url: offerUrl },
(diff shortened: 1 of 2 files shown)


affiliate/blocks/productCard/insertLink/InsertAffiliateLink.jsx
```diff
@@ -35,8 +35,9 @@ const InsertAffiliateLink = ({
     () => setShowModal(true),
     [setShowModal],
   );
-  const onProductSelect = useCallback((product, offerUrl) => {
+  const onProductSelect = useCallback((productInfo) => {
     const { start, end } = value;
+    const { offerUrl, product } = productInfo;
```
amittman, line 40: Two small things:
1. Can we add a comment here and in the other blocks' Edit pages regarding where we can find the "message" being sent from the platform? Something like: `// format of arguments is based on data received from postMessage event in orgnc/platform file: IntegrationProductSearch.jsx` (the product carousel file is `IntegrationProductCarousel.tsx`). We will have to change this when making updates to the file naming but it will make debugging a bit easier when coming back to this later on.
2. Can we slightly modify the function argument to deconstruct the arg from the beginning, like we see in the other `on{...}Select` calls in the `Edit` files?
```js
const onProductSelect = useCallback({ product, offerUrl }) => {
  // ...
};
```

amittman (approved):
LGTM, thanks!

affiliate/blocks/productCard/insertLink/InsertAffiliateLink.jsx
Mwindo: Good thoughts. Will do.

Mwindo: Waiting on ProductCard since that will be refactored soon.

PR #137, opened 2023-01-05
EP-4447: Refactor affiliate product card to use iframed product card creation widget

Mwindo:
This PR modifies the product card so that it is built using the platform widget's component rather than using separate, WordPress-specific form controls. This ensures that any changes we make to product card creation can be executed in one spot rather than having to make the changes both in the platform code (widgets page) and in the wordpress-plugin code. The PR also changes the edit icon for both the product card and the product carousel from a magnifying glass to a pencil.

Note that as soon as this PR is in production (when this PR is released and the solutions-wordpress code is updated to use the latest wordpress-plugin version), creating new product cards will be temporarily broken until https://github.com/orgnc/organic-web/pull/1486 is released. Once this organic-web PR is released, existing product cards will be temporarily broken, but product creation will work. See the organic-web PR for details.

Before:

https://user-images.githubusercontent.com/47676832/210881223-616b8ed4-170d-46f3-b04f-56e73d7da204.mov

After:

https://user-images.githubusercontent.com/47676832/210880714-b5967eb5-255f-4392-8893-da2801aeb513.mov

### DIFF
--- a/Organic/Affiliate.php
+++ b/Organic/Affiliate.php
@@ -31,12 +31,14 @@ class Affiliate {
             $this->organic->version
         );
         $product_search_page_url = $this->organic->getPlatformUrl() . '/apps/affiliate/integrations/product-search';
+        $product_card_creation_url = $this->organic->getPlatformUrl() . '/apps/affiliate/integrations/product-card';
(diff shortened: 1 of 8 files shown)


amittman (approved):
LGTM, thanks @Mwindo

PR #138, opened 2023-01-06
EP-3440: Remove "breaks ads" text from WP plugin settings

tonystanhope:
Quick single-line cleanup. There's no indication that V2 breaks ads, but this was added early during testing and had not yet been cleaned up.


NitishMudgal (approved):
LGTM

PR #140, opened 2023-01-09
Bugfix: Fix ProductCard Save.jsx

Mwindo:


affiliate/blocks/productCard/src/Save.jsx
```diff
@@ -1,14 +1,13 @@
 import { useBlockProps } from '@wordpress/block-editor';
 
-import ProductCarousel from './ProductCard';
```
Mwindo, line 3: Well, this certainly won't do. I modified this file and forgot to push the change in my previous PR.

Mwindo:
Going to merge and release this since it fixes something clearly broken, but happy to get a review afterwards.

affiliate/blocks/productCard/src/Save.jsx
Mwindo: I've been testing on a feature branch, so nothing is broken in production.

amittman (commented):
Ah, looks good. Unfortunate we didn't catch this with the previous. Should be able to set up tests for this branch this sprint.

Mwindo:
> Should be able to set up tests for this branch this sprint.

I was thinking the same thing. It will be super helpful.

PR #141, opened 2023-01-17
EP-2249: Plugin Refactoring To Meet Wordpress.Org Requirements

iceone:


sergunich (approved):
Looks good 👍

build-zip.sh
```diff
@@ -1,20 +1,38 @@
...
+
+# cp -r vendor build/organic/
 (cd affiliate/ && npm ci && npm run build)
-mkdir -p build/organic/affiliate/
-cp -r affiliate/blocks/ build/organic/affiliate/
+mkdir -p build/organic/affiliate/productCard build/organic/affiliate/productCarousel
+cp -r affiliate/blocks/productCard/build build/organic/affiliate/productCard/
+cp affiliate/blocks/productCard/block.json build/organic/affiliate/productCard/
+cp -r affiliate/blocks/productCarousel/build build/organic/affiliate/productCarousel/
+cp affiliate/blocks/productCarousel/block.json build/organic/affiliate/productCarousel/
```
Mwindo, line 35: I don't think this will work. See https://organicly.atlassian.net/browse/EP-2249?focusedCommentId=58233. We register scripts based on the affiliate/blocks path [here](https://github.com/orgnc/wordpress-plugin/blob/9d1afb7bd0932f7f69f8b0412858c56862bd2fac/Organic/Affiliate.php#L21-L32), which looks like it wasn't changed in this PR. Is there a reason we need to modify this portion of the build file?

PR #142, opened 2023-01-18
Run script for affiliate SDK to render widgets blocks on admin page load

Mwindo:
EDIT: I can just use the [load](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event) event 🤦. I had a pretty hacky fix in place before.

In order to properly display existing affiliate widget blocks in the WordPress Post editor, we need to run our affiliate JavaScript SDK (once, ideally) after the widgets blocks are all loaded.

See https://organicventures.slack.com/archives/C01LLSNMKT2/p1674149355631459.

Before (SDK isn't run on Post load at the appropriate time, so widgets don't display properly):

Screen Shot 2023-01-18 at 3 20 19 PM

After:

https://user-images.githubusercontent.com/47676832/213286374-697e0bb8-1f87-416c-af66-dc62f5befab8.mov


affiliate/initSDKOnPostLoad.js
```diff
@@ -0,0 +1,20 @@
...
+    const integrations = document.querySelectorAll('[data-organic-affiliate-integration]');
+    if (integrations.length > 0) {
+      window.empire?.apps?.affiliate?.init?.();
+    }
+  } else {
+    setTimeout(() => waitForPageLoadAndInitSDK(newBlocks, remainingTries - 1), 500);
+  }
+};
+
+waitForPageLoadAndInitSDK([], 20);
```
Mwindo, line 20: Limit (arbitrarily) to 10 seconds. I'm not sure if this is necessary, but in case something goes wrong, we don't want this to be running indefinitly.

affiliate/initSDKOnPostLoad.js
```diff
@@ -0,0 +1,20 @@
...
+ */
+const waitForPageLoadAndInitSDK = (oldBlocks, remainingTries) => {
+  if (remainingTries < 1) {
+    return;
+  }
+  const newBlocks = document.querySelectorAll('[data-block]');
+  if (newBlocks.length > 0 && newBlocks.length === oldBlocks.length) {
+    const integrations = document.querySelectorAll('[data-organic-affiliate-integration]');
+    if (integrations.length > 0) {
+      window.empire?.apps?.affiliate?.init?.();
```
sergunich, line 13: Actually, SDKv2 will be auto-initialized on itself unless auto-initialization is explicitly disabled
https://github.com/orgnc/platform/blob/81e168049dbdbacb313e9e1831035e690c3941f5/frontend/sdk/src/sdkv2.js#L39

So, instead, to make sure it's correctly initialized and the page processed even if SDKv2 loads after all blocks are loaded we should use `cmd` queue
```js
    window.empire.cmd.push(({ affiliate, core }) => {
        // Core
        if (core && !core.isInitialized()) {
          core.init();
        }

        // Affiliate
        if (affiliate && affiliate?.isEnabled()) {
          if (!affiliate.isInitialized()) {
            affiliate.init();
          } else {
            affiliate.processPage();
          }
        }
    }
```

Mwindo: Ah yes, that makes sense. Thanks Sergey.

aprietof (commented):
I left a couple of suggestions but it looks good to me.

Organic/Affiliate.php
```diff
@@ -58,4 +59,17 @@ public function register_gutenberg_block() {
...
+        // Scripts to enqueue only for WP Post pages.
+        if ( 'post.php' != $hook_suffix && 'post-new.php' != $hook_suffix ) {
+            return;
+        }
+        wp_enqueue_script(
+            'on-post-load-scripts',
+            plugins_url( 'affiliate/initSDKOnPostLoad.js', __DIR__ ),
+            [ 'organic-sdk' ],
+            $this->organic->version
+        );
```
aprietof, line 73: The code looks functional to me, I have a couple of suggestions tho:

In my experience using if statements that return truthy values instead of falsy values perhaps makes the code more readable and self-explanatory, it helps make clear what you expect to be true. 

Also, I would suggest the use of `===` instead of `==` which can prevent unexpected behavior and bugs caused by type conversion as a general practice. What do you think? @Mwindo 

```php
if ( 'post.php' === $hook_suffix && 'post-new.php' === $hook_suffix ) {
  wp_enqueue_script(
      'on-post-load-scripts',
      plugins_url( 'affiliate/initSDKOnPostLoad.js', __DIR__ ),
      [ 'organic-sdk' ],
      $this->organic->version
  );
}
```

aprietof (commented):
I left a couple of suggestions but it looks good to me.

Organic/Affiliate.php
Mwindo: Yes, this is better. I'm not sure why I didn't write it this way originally. Thank you!

Organic/Affiliate.php
```diff
@@ -12,6 +12,7 @@ public function __construct( Organic $organic ) {
         $this->organic = $organic;
         add_action( 'init', [ $this, 'register_gutenberg_block' ] );
         add_action( 'admin_enqueue_scripts', [ $this, 'register_scripts' ] );
+        add_action( 'admin_enqueue_scripts', [ $this, 'admin_post_page_enqueue' ] );
```
iceone, line 15: I'd prefer to enqueue all the scripts in the single `register_scripts` method rather that having multiple hooks

iceone: Also, I think it would be nice to wrap admin-related stuff/hook registration into `is_admin()` condition.

iceone (approved):

PR #144, opened 2023-01-24
Modify build-zip.sh to include new affiliate SDK script

Mwindo:
In https://github.com/orgnc/wordpress-plugin/pull/142, I added a new affiliate script that would run our SDK so that widgets displayed correctly. I tested locally, but testing on a feature PR [here](https://github.com/orgnc/solutions-wordpress/pull/3837), I realized that build-zip.sh hadn't been updated to copy the file containing the script (which, no surprises, leads to a 404 error in the console). This PR fixes the issue. I've tested locally once more, including a test with the plugin directory built via build-zip.sh to better mimic the feature/production environment.


iceone (approved):

aprietof (commented):
👍

PR #161, opened 2023-04-06
EP-5061: Add affiliate block script dependencies to wp_register_script

Mwindo:
I should have named this branch `bugfix/...` but I didn't think it was worth creating a new branch.

In any case, this PR fixes an error in which affiliate blocks were not working properly because their dependencies were not registered correctly. See https://github.com/WordPress/gutenberg/blob/trunk/docs/how-to-guides/javascript/js-build-setup.md#dependency-management.

Quick video showing them working again (ignore the awful styling--I was testing something else locally):

https://user-images.githubusercontent.com/47676832/230426534-446f91bf-003c-43f0-9ca2-0c949e323295.mov

### DIFF
--- a/src/Organic/Affiliate.php
+++ b/src/Organic/Affiliate.php
@@ -25,16 +25,19 @@ class Affiliate {
             // Gutenberg blocks are not supported
             return;
         }
+        // As more blocks are added, create a function like "register_widget_block_script".
+        $card_asset_file = include( plugin_dir_path( __DIR__ ) . 'blocks/affiliate/productCard/build/index.asset.php' );
         wp_register_script(
             'organic-affiliate-product-card',
             plugins_url( 'blocks/affiliate/productCard/build/index.js', __DIR__ ),
-            [ 'organic-sdk' ],
+            $card_asset_file['dependencies'],
             $this->organic->version
         );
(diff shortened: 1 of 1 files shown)


src/Organic/Affiliate.php
```diff
@@ -25,16 +25,22 @@ public function register_scripts( $hook_suffix ) {
             // Gutenberg blocks are not supported
             return;
         }
+        // As more blocks are added, create a function like "register_widget_block_script".
+        $card_asset_file = include( plugin_dir_path( __DIR__ ) . 'blocks/affiliate/productCard/build/index.asset.php' );
+        // wp-scripts 5.0.0+ generates an index.asset.php file with dependencies. If the wp-scripts version is before
```
sergunich, line 30: Do we really need this legacy handling?

The `index.asset.php` is generated by our scripts that use the latest `wp-scripts` version. So, the `index.asset.php` will be generated and present in the plugin and accessible to read even if the plugin is installed on the older version of the WP

sergunich (approved):
Looks good 👍

src/Organic/Affiliate.php
Mwindo: Ah, good point. I'll get rid of this then. Thanks!

PR #162, opened 2023-04-13
EP-5058: WP Plugin Selenium

Mwindo:
This PR:

- Modifies the dev environment to use `http://wpplugin.lcl.organic.ly[:port]` rather than `localhost:[port]`.`*`
- Adds initial selenium setup and a couple basic affiliate tests.

**Note: I will squash commits.**

`*` Why? WordPress routes to `localhost` by default, which docker services cannot access without some less-than-ideal workarounds (like running [host network mode](https://docs.docker.com/network/host/), only supported on Linux). Instead, we need to have our URL accessible in both the host and the docker environments. We do this by setting `WP_HOME`/`WP_SITEURL` to URL X (so the host can access X), aliasing our services with X (so the container can access X), and finessing the ports so we can use `wpplugin.lcl.organic.ly[:port]` rather than `wpplugin-[some-service].lcl.organic.ly` (which would entail separate entries in the local `etc/hosts` file for every WP service we might want).

TODO in [follow-up PR](https://organicly.atlassian.net/browse/EP-5160):

- Run Selenium tests as a separate (self-hosted?) GitHub actions workflow.

Quick video showing the routing:

https://user-images.githubusercontent.com/47676832/231800817-b7b6549a-e158-48bf-9dcd-25243ad0bc42.mov


docker-compose.yml
```diff
@@ -109,12 +142,44 @@ services:
       <<: *base-build
       target: builder
     # on fresh build/start node_modules will be empty until 'npm install'
-    command: bash -c "until npx nodemon --ignore "*/build/*.js"; do sleep 5; done"
+    command: bash -c "npm install; npm run build"
```
Mwindo, line 145: Note to self: Undo this.

src/composer.json
```diff
@@ -3,15 +3,16 @@
     "description": "Wordpress Plugin for Organic Platform",
     "type": "wordpress-plugin",
     "require": {
-        "gmostafa/php-graphql-client": "^1.10",
```
Mwindo, line 6: Note to self: undo insofar as possible.

README.md
```diff
@@ -32,6 +32,10 @@ $ cp .env.example .env
 ```
 
 7. Don't forget to update `TODO` vars in your `.env` file!
+8. Make sure your `/etc/hosts` file contains `wpplugin.lcl.organic.ly`.
+```bash
+$ organic dev update-hosts
```
Mwindo, line 37: If/once the approach in this PR is approved, I'll open and merge PR for the update to the hosts.yaml file in organic dev.

dev.py
```diff
@@ -13,7 +13,7 @@
 
 ADMIN_USER = 'organic'
 ADMIN_PASSWORD = 'organic'
-DEFAULT_WP_SERVICE = 'wp61-php74'
+DEFAULT_WP_SERVICE = 'wp61-php74'  # 'wordpress.lcl.default.com'
```
Mwindo, line 16: Note to self: remove

dev.py
```diff
@@ -305,6 +305,14 @@ def lint(filenames, php, js):
         service_trigger('nodejs', 'npm run lint:js')
 
 
+@cli.command()
+def run_tests():
+    if _service_is_running('composer'):
+        service_trigger('composer', './vendor/bin/phpunit')
+    else:
+        info('Need composer service to be running. Have you run the up command?')
```
Mwindo, line 313: The composer service needs to be running in order to run tests. Have you run the up command?

dev.py
```diff
@@ -305,6 +312,42 @@ def lint(filenames, php, js):
...
+            )
+            return
+
+    for service in services:
+        if not _service_is_running(service):
+            info(f"Cannot run tests for {service}: {service} is not running.", "red")
+        else:
+            port = config.get_service_port(service)
+            version = config.get_wp_version(service)
+            print(version)
```
Mwindo, line 343: Remove

src/tests/test_affiliate/WidgetsTest.php
```diff
@@ -0,0 +1,63 @@
...
+        fwrite(STDERR, print_r('\n Getting Selenium Browser', TRUE));
+        $browser = SeleniumBrowser::get_test_browser();
+        try {
+            fwrite(STDERR, print_r('\n Going to page', TRUE));
+            $browser->go_to_new_post();
+            $browser->add_block( $block_type );
+            $browser->wait( .5 );
+            $url_correct = str_contains( $browser->get_iframe_url( 0 ), 'app.organic.ly');
+            $browser->quit();
+            $this->assertTrue( $url_correct );
```
Mwindo, line 39: For reasons I haven't yet identified, WP 5.9 fails here and says the iframe src is about:blank.

docker-compose.yml
```diff
@@ -115,6 +147,34 @@ services:
...
+      # https://github.com/seleniumhq-community/docker-seleniarm#running-the-arm-container-images
+      # Visit http://localhost:7901 to see what's happening in the browser
+      # during selenium tests
+      #
+      # !!! VNC password is "secret" !!!
+      - 5901:5900  # port for external VNC client
+      - 7901:7900  # noVNC.com UI
+
+  nginx-proxy:
+    image: nginxproxy/nginx-proxy:1642
```
Mwindo, line 167: I used the same image as in organic-dev for consistency.

docker-context/Dockerfile
```diff
@@ -30,25 +30,37 @@ RUN { \
 		echo 'memory_limit = 1024M'; \
 	} > /usr/local/etc/php/conf.d/tap.ini;
 RUN wp --allow-root package install https://github.com/nlemoine/wp-cli-fixtures/archive/refs/tags/0.3.0.zip
+COPY apache_port_template.conf /etc/apache2/ports.conf
+# Make sure the port override below matches the docker-compose port for wp49-php72
+RUN sed -i 's/<PORT>/8031/g' /etc/apache2/ports.conf
```
Mwindo, line 35: To avoid having a separate URL for each WP service, we tell each WP service to listen to the ports we define in our docker-compose file.

src/tests/SeleniumBrowser.php
```diff
@@ -0,0 +1,283 @@
+<?php
```
Mwindo, line 1: Based a lot upon our platform [BrowserHelper](https://github.com/orgnc/platform/blob/c880cf2da990b05b8fd1472eb8b47f35d640a63e/backend/tests/base.py#L155).

src/tests/SeleniumBrowser.php
```diff
@@ -0,0 +1,283 @@
...
+        $this->click( '.editor-block-list-item-' . $block_type );
+    }
+
+    /**
+     * @param int $index
+     * @return mixed
+     * @throws Exception
+     */
+    function get_iframe_url( int $index ) {
+        $iframe = $this->wait_for('iframe', null, null, true)[$index];
```
Mwindo, line 278: I'll fix spacing here in next commit.

src/tests/test_affiliate/WidgetsTest.php
```diff
@@ -0,0 +1,68 @@
...
+define( "Organic\WP_VERSION", getenv( 'WP_VERSION' ) ?? '' );
+
+class WidgetsTest extends TestCase {
+
+    /**
+     * Tests for custom widgets.
+     * Note that these tests will fail for non-Gutenberg WordPress (< version 5).
+     */
+
+    function wordPressVersionTooLow() : bool {
```
Mwindo, line 19: The format in this file is mixed up, but I'm not sure exactly what we want to do since we tend to use snake_case for our own code but other PHP libraries use camelCase.

src/tests/test_affiliate/WidgetsTest.php
```diff
@@ -0,0 +1,68 @@
...
+            // some other WP versions). The page content looks basically the same as for 6.1,
+            // which works fine, and I can visually see the correct URL in Selenium Grid's live view.
+            // Let's get rid of this when we figure out the issue.
+            if ( !empty( WP_VERSION ) && WP_VERSION == '5.9' ) {
+                $url_correct = true;
+            } else {
+                $url_correct = str_contains( $browser->get_iframe_url( 0 ), 'app.organic.ly' );
+            }
+            $browser->quit();
+            $this->assertTrue( $url_correct );
```
Mwindo, line 45: I spent several hours on this and didn't want to waste any more time. I can revisit later. Also happy to entertain any suggestions!

src/tests/SeleniumBrowser.php
Mwindo: I'm not _thrilled_ about some of the selectors I've used for WP-specific elements, but since I don't have control over what WordPress does, I did my best.

src/tests/SeleniumBrowser.php
```diff
@@ -0,0 +1,282 @@
...
+     * @throws Exception
+     */
+    function add_block( string $block_type ) {
+        // First, click the first Add block button.
+        $this->click( '[aria-label="Toggle block inserter"]' );
+        // Click the block search bar. Search for the block type.
+        $this->fill_text_input( '.components-search-control__input', $block_type );
+        // Click the icon for the block type to insert it.
+        $this->click( '.editor-block-list-item-' . $block_type );
+    }
```
Mwindo, line 270: In general, version 5.9 seems flakier than version 6. There is an Add block button in the paragraph editor (first image below), but if I click that in 5.9, it sort of just disappears and then reappears? I've chosen to click the toolbar button (second image below) instead.

Screen Shot 2023-04-14 at 5 34 24 PM

Screen Shot 2023-04-14 at 5 34 28 PM

docker-compose.yml
Mwindo: Although I didn't do this for the selenium image. I thought having an updated selenium was more important.

.github/workflows/code-quality.yaml
```diff
@@ -17,12 +17,13 @@ jobs:
...
           working_dir: src
           php_version: "7.4"
+          php_extensions: zip
 
       - name: Run PHPCS
         run: cd src && composer run lint
 
       - name: Run PHPUnit
-        run: cd src && composer run phpunit
+        run: cd src && composer run phpunit -- --exclude-group selenium_test
```
Mwindo, line 26: In the follow-up PR, I'm thinking we can have a separate action to run tests (both selenium and non-selenium). The matrix strategy we use [here](https://github.com/orgnc/platform/blob/c880cf2da990b05b8fd1472eb8b47f35d640a63e/.github/workflows/pr-tests.yaml#L110) is probably overkill, but I think separating Code Quality from Tests is probably a good idea. What do you think @sergunich?

src/composer.json
Mwindo: @sergunich: Nevermind, this seems to be working. I'll revert these versions. That said, ./dev up is inconsistent for me. Sometimes the command breaks somewhere, which leads to the composer not installing updates and the WP environment not being set up (hence my original difficulties accessing the database in this PR). I'll hopefully identify the issue and fix it in this sprint.

.github/workflows/code-quality.yaml
```diff
@@ -17,12 +17,13 @@ jobs:
         with:
           working_dir: src
           php_version: "7.4"
+          php_extensions: zip
```
Mwindo, line 20: Required by php-webdriver

src/tests/test_affiliate/WidgetsTest.php
sergunich: Actually we do use camelCase in the Organic WP plugin, the only snake_case code is methods provided by Wordpress itself

docker-context/Dockerfile
```diff
@@ -30,25 +30,37 @@ RUN { \
...
 COPY . /tmp/docker-context
 
 FROM wordpress:5.9-php7.4 AS wp59-php74
 COPY --from=wordpress:cli-php7.4 /usr/local/bin/wp /usr/local/bin/wp
 COPY --from=builder /root/.wp-cli /root/.wp-cli
 COPY --from=builder /tmp/base-theme.zip /tmp/base-theme.zip
 COPY --from=builder /tmp/wpamp-plugin.zip /tmp/wpamp-plugin.zip
+COPY apache_port_template.conf /etc/apache2/ports.conf
+# Make sure the port override below matches the docker-compose port for wp59-php74
```
sergunich, line 44: Probably worth to define these as `WP59_PHP74_PORT` in `.env` file and reuse it in `Dockerfile` and `docker-compose.yml`

docker-compose.yml
```diff
@@ -115,6 +147,34 @@ services:
       - ./src/blocks:/blocks
       - blocks-node-modules:/blocks/node_modules
 
+  selenium:
+    image: ${SELENIUM_IMAGE:-selenium/standalone-chrome:4.8.3}
+    # image: ${SELENIUM_IMAGE:-seleniarm/standalone-chromium:4.8.3} # use this for ARM64
```
sergunich, line 152: I think this will be nicer instead of commented `image:`
```
# For ARM64 set `SELENIUM_IMAGE=seleniarm/standalone-chromium:4.0.0-20211213` in your `.env` file
```

docker-compose.yml
```diff
@@ -115,6 +147,34 @@ services:
...
+      - 7901:7900  # noVNC.com UI
+
+  nginx-proxy:
+    image: nginxproxy/nginx-proxy:1642
+    ports:
+      - "80:80"
+      - "443:443"
+    volumes:
+      - /var/run/docker.sock:/tmp/docker.sock:ro
+      # - ./config/nginx.conf:/etc/nginx/nginx.conf:delegated
```
sergunich, line 173: Can be deleted?

sergunich (approved):
Looks good! These wil be very valuable for stability of the plugin and new tests 👍

dev.py
```diff
@@ -305,6 +312,46 @@ def lint(filenames, php, js):
...
+@click.pass_obj
+def run_tests(config, services, exclude):
+
+    if not _service_is_running("composer"):
+        info(
+            "The composer service needs to be running in order to run tests. "
+            "Have you run the up command?",
+            "red",
+        )
+        return
```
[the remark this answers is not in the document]
Mwindo: Good call, thank you!

dev.py
```diff
@@ -233,7 +240,7 @@ def up(config, services, build, reset, install_amp, pull_configs):
     up_cmd.append('--force-recreate')
     # Remove orphaned containers after docker-compose.yml edits
     up_cmd.append('--remove-orphans')
-    up_cmd.extend(services)
+    up_cmd.extend([*services, 'nginx-proxy'])
```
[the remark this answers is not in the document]
Mwindo: Good catch, I think so. (I had it this way because of something in my original setup causing problems.)

docker-compose.yml
```diff
@@ -53,10 +55,11 @@ services:
     build:
       <<: *base-build
       target: wp49-php72
+      args:
+        - WP49_PHP72_PORT=${WP49_PHP72_PORT}
     <<: *base-wordpress
     ports:
-      # Make sure this port matches the Dockerfile port override for wp49-php72.
-      - 8031:8031
+      - ${WP49_PHP72_PORT}:${WP49_PHP72_PORT}
```
sergunich, line 60: @Mwindo, looks like you forgot to update the `.env.template` and we should add these env vars to the REQUIRED_ENVVARS list
https://github.com/orgnc/wordpress-plugin/blob/master/dev.py#L18

Mwindo: Good catch, thanks!

Mwindo: Will do that now.

Mwindo: https://github.com/orgnc/wordpress-plugin/pull/165

PR #163, opened 2023-04-17
Enqueue site custom css

PetrMitsel:


src/Organic/PageInjection.php
```diff
@@ -193,6 +194,12 @@ public function injectBrowserSDK() {
         }
     }
 
+    public function injectOrganicCustomCss() {
+        $siteId = $this->organic->getSiteId();
+        $cssUrl = $this->organic->getRestAPIUrl() . '/sdk/customcss/' . $siteId;
+        wp_enqueue_style( $siteId, $cssUrl, [], '1.0' );
```
PetrMitsel, line 200: Not sure what to do with this version number, open to suggestions.

src/Organic/Organic.php
```diff
@@ -23,6 +23,7 @@ class Organic {
     public $version = \Organic\ORGANIC_PLUGIN_VERSION;
 
     const DEFAULT_PLATFORM_URL = 'https://app.organic.ly';
+    const DEFAULT_REST_API_URL = 'https://api.organic.ly';
```
Mwindo, line 26: Is there any reason we need to specify REST?

src/Organic/PageInjection.php
Mwindo: I guess the question is how important is cache invalidation for this CSS? If, as I suspect, it is important, then we probably need to do something else. Maybe @sergunich has better ideas, but here are a couple options:
- Using a `custom_css_version` field (an int, or a short timestamp) that updates whenever the `custom_css` field is updated.
- Having an `sdk_version` field that updates whenever the sdk config is changed. (More general than the above, but also necessarily less granular. I don't know how often sdk config changes are made; if frequently, then the above is probably better.)

src/Organic/Organic.php
PetrMitsel: Technically no, but the environment variable `ORGANIC_API_URL` currently points to the graphql url I believe, so I named it that to avoid confusion.

Mwindo: I see. If we're phasing out the REST URL anyway, maybe we can just use the `ORGANIC_API_URL` as is. But I don't have strong opinions here.

src/Organic/PageInjection.php
Mwindo: And the first discussion post [here](https://developer.wordpress.org/reference/functions/wp_enqueue_style/) lists a solution as well.

PetrMitsel: @Mwindo I'm not sure the solution in the discussion post will work for us because we'd need last modified time for a remote file (technically just a text response). There are some implementations of this, but they seem overly complicated compared to your other proposals.

Mwindo: I just realized I didn't see this other PR. It looks like maybe caching was already taken care of (similar to my suggestion in part 2). https://github.com/orgnc/platform/pull/3240/files#r1161824120. If so, it doesn't really matter what version we pass in I don't think.

src/Organic/Organic.php
sergunich: We would still need REST API for things like [ads.txt and sellers.json](https://github.com/orgnc/platform/blob/master/backend/ads/api/rest.py#L41-L79)

But in this case, we should use `https://api.organiccdn.io/` instead
https://organicly.atlassian.net/wiki/spaces/NTE/pages/3080361/Fastly+setup+for+organiccdn.io+and+app.organic.ly

src/Organic/PageInjection.php
sergunich: Yes, we don't need any special cache handling here because it will work similar to SDK serving (cache and invalidation handled by Fastly)

PR #165, opened 2023-04-20
Bugfix: Update .env.template and required vars

Mwindo:
See https://github.com/orgnc/wordpress-plugin/pull/162#discussion_r1172993316.


sergunich (approved):
👍

PR #166, opened 2023-04-21
EP-5059: Sentry integration

Mwindo:
See https://github.com/orgnc/platform/pull/3272 for the related platform code. I still need to address [this comment](https://github.com/orgnc/wordpress-plugin/pull/166#discussion_r1174122481), but I'd like some more feedback in the meantime.

In this PR:

- Unless the publisher has chosen to disable the option, we set up an Organic Sentry client with a default DSN to capture errors.
- If we have a site-specific Sentry DSN on the platform (which we sync to the WP database intermittently), we reconfigure the Organic Sentry client to us that instead.

The new publisher option to disable Sentry:
Screen Shot 2023-04-21 at 6 55 40 PM


src/Organic/Organic.php
```diff
@@ -3,7 +3,6 @@
 namespace Organic;
 
 use DateTime;
-use GuzzleHttp\Exception\GuzzleException;
```
Mwindo, line 6: This doesn't seem to be used at all.

src/Organic/Organic.php
```diff
@@ -14,7 +13,9 @@
 const CAMPAIGN_ASSET_META_KEY = 'empire_campaign_asset_guid';
 const GAM_ID_META_KEY = 'empire_gam_id';
 const SYNC_META_KEY = 'empire_sync';
-
+// The DSN to use before we load client-specific DSNs.
+const DEFAULT_SENTRY_DSN = 'https://e1cf660e5b3947a4bdf7c516afaaa7d2@o472819.ingest.sentry.io/4505048050434048';
```
Mwindo, line 17: Sentry says it's "OK" for DSN keys to be public. (See https://docs.sentry.io/product/sentry-basics/dsn-explainer/.) This seems questionable to me. I'll deactivate this default and create a new one. We should pass in the default as a secret upon build.

src/Organic/SDK/OrganicSdk.php
```diff
@@ -23,7 +23,7 @@
  */
 class OrganicSdk {
 
-    const DEFAULT_API_URL = 'https://api.organic.ly/graphql';
+    const DEFAULT_API_URL = 'http://api.lcl.organic.ly:8000/graphql';
```
Mwindo, line 26: Whoops, remove

src/composer.json
```diff
@@ -58,10 +59,11 @@
     "config": {
         "allow-plugins": {
             "dealerdirect/phpcodesniffer-composer-installer": true,
-            "liborm85/composer-vendor-cleaner": true
+            "liborm85/composer-vendor-cleaner": true,
+            "php-http/discovery": true
         },
         "platform": {
-            "php": "7.2"
+            "php": "7.3"
```
Mwindo, line 66: Fix this, and figure out why composer causes such headaches.

src/Organic/Organic.php
```diff
@@ -15,6 +14,9 @@
 const GAM_ID_META_KEY = 'empire_gam_id';
 const SYNC_META_KEY = 'empire_sync';
 
+// The DSN to use before we load client-specific DSNs.
+const DEFAULT_SENTRY_DSN = 'https://e1cf660e5b3947a4bdf7c516afaaa7d2@o472819.ingest.sentry.io/4505048050434048';
```
Mwindo, line 18: I don't like having this public, but 1) Sentry (questionably) [says it's ok](https://docs.sentry.io/product/sentry-basics/dsn-explainer/#dsn-utilization) and 2) I don't see a way around it unless we provide the DSN in the platform and have them paste it into their WordPress plugin settings (like we do for the API key).

Mwindo: Another option: we only get the DSN from the platform, which either returns the client-specific DSN or else the some default DSN. The only disadvantage here is that we don't get any Sentry logging for initializing the plugin.

Mwindo: Another option: we only get the DSN from the platform, which either returns the client-specific DSN or else the some default DSN. The only disadvantage here is that we don't get any Sentry logging for initializing the plugin.

src/Organic/Organic.php
```diff
@@ -232,6 +233,12 @@ public function init( $apiUrl = null, $cdnUrl = null ) {
         $this->adsTxt = new AdsTxt( $this );
 
         $this->isEnabled = $this->getOption( 'organic::enabled' );
+        $this->logToSentry = $this->getOption( 'organic::log_to_sentry' );
+        // Initialize Sentry as soon as possible.
+        if ( $this->isEnabled && $this->logToSentry ) {
+            $this->initSentry( $this->sdk );
+        }
```
Mwindo, line 240: Originally I initialized Sentry in `organic.php` with a default DSN to catch errors during the initialization of Organic and OrganicSDK instances even before setting a client-specific DSN. However, I don't think we can do that if we rely on an Admin Settings option.

sergunich: You can use `get_option` directly to check if we should initialize Sentry with default DSN
https://developer.wordpress.org/reference/functions/get_option/

We should skip this initialization only if `organic::log_to_sentry` is explicitly set to false

src/Organic/Organic.php
```diff
@@ -277,6 +284,31 @@ public function init( $apiUrl = null, $cdnUrl = null ) {
...
 
+    /**
+     * @param OrganicSdk $sdk
+     */
+    public function initSentry( \Organic\SDK\OrganicSdk $sdk ) {
+        if ( $this->environment != 'PRODUCTION' ) {
+            return;
+        }
+        try {
+            $config = $sdk->queryWordPressConfig();
```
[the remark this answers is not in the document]
Mwindo: Got it. Thank you!

src/Organic/Organic.php
sergunich: Usually - yes, it should be done through an Organic wrapper, but here we have a special case when we need an option before we have access to the wrapper

src/Organic/Organic.php
```diff
@@ -277,6 +284,31 @@ public function init( $apiUrl = null, $cdnUrl = null ) {
...
+        } catch ( \Exception $e ) {
+            self::captureException( $e );
+        }
+        if ( ! empty( $config ) ) {
+            $sentryDSN = $config['sentryDsn'];
+            if ( ! empty( $sentryDSN ) ) {
+                \Sentry\init(
```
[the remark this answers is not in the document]
Mwindo: > This will overwrite the global setup of Sentry for the current site

Ah, duh, I see 🤦 . This clarifies why references to Sentry were in the codebase without it being a requirement. I (obviously) didn't think of the interaction with any existing Sentry on the client WordPress. Thank you!

src/Organic/Organic.php
```diff
@@ -277,6 +284,31 @@ public function init( $apiUrl = null, $cdnUrl = null ) {
...
+     * @param OrganicSdk $sdk
+     */
+    public function initSentry( \Organic\SDK\OrganicSdk $sdk ) {
+        if ( $this->environment != 'PRODUCTION' ) {
+            return;
+        }
+        try {
+            $config = $sdk->queryWordPressConfig();
+        } catch ( \Exception $e ) {
+            self::captureException( $e );
```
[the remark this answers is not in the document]
Mwindo: `self` references the class and `this` references the "current" instance of the class; in this case, since `captureException` is a static method (a method on the class itself), I believe we cannot use `this`. (It isn't even an option in my IDE, for instance.) I could use `static` instead of `self`.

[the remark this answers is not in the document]
Mwindo: I see. It looks like I can also use `$this::captureException`. From what I can tell, though, this is a nonstandard usage of $this.

src/Organic/Organic.php
```diff
@@ -1285,12 +1339,18 @@ public function getSplitTestPercent(): int {
     }
 
     public static function captureException( \Exception $e ) {
-        if ( function_exists( '\Sentry\captureException' ) ) {
```
Mwindo, line 1288: This should always be true now.

src/organic.php
```diff
@@ -12,14 +12,34 @@
 
 use Organic\Organic;
 
+// The DSN to use before we load client-specific DSNs.
+const DEFAULT_SENTRY_DSN = 'https://e1cf660e5b3947a4bdf7c516afaaa7d2@o472819.ingest.sentry.io/4505048050434048';
```
Mwindo, line 16: Sentry (somewhat questionably, I think) says that making the DSN public is ok (see [here](https://docs.sentry.io/product/sentry-basics/dsn-explainer/#dsn-utilization)). I'd prefer not to hard-code this, although I don't know how to do that without having the client copy-and-pasting a DSN, which seems even worse (tedious and error-prone).

src/Organic/SimpleConfigSyncCommand.php
```diff
@@ -0,0 +1,64 @@
+<?php
```
Mwindo, line 1: Just pulling out some duplicate code into a helper class. `AdsTxtSyncCommand` is very similar but not _quite_ the same, so I left it alone for now.

sergunich: Awesome 👍

src/organic.php
sergunich: Looks good! 
It's okay for DSN to be public. In the case of frontend (Platform UI or Organic SDK), there is no other way except hardcoding into a bundle - so, here it's the same 🙂

sergunich (approved):
Looks good! 👍

src/Organic/Organic.php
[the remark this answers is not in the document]
sergunich: Ohh, I see... Yeah, makes sense!

src/composer.json
Mwindo: @sergunich Do we need this to be 7.2 instead of 7.3? I'm getting composer conflicts unrelated to any requirements I've added whenever I do a composer update. (The conflicts have to do with symfony and the php version.)

PR #167, opened 2023-04-24
EP-5160: Selenium in GitHub Actions

Mwindo:
This PR runs the Selenium tests upon opening a pull request.

When the PR is approved, I will squash the messy commits.

---

Old comments:
I'm keeping this as a draft for now since I am seeing some flakiness in the build process that I need to figure out. Examples:
- Composer auth issue: https://github.com/orgnc/wordpress-plugin/actions/runs/4800732577/jobs/8542037898. Maybe it is ok now that I've changed how I pass in the COMPOSER_AUTH variable (perhaps I was setting it incorrectly), but I need to check. **EDIT: I've added a composer config line that seems to solve this.**
- Unhealthy DB happened once: https://github.com/orgnc/wordpress-plugin/actions/runs/4801639382/jobs/8544121786. Maybe affects local dev too. **EDIT: I've increased the retry count from 3 to 5.**


.github/workflows/code-quality.yaml
```diff
@@ -27,3 +27,29 @@ jobs:
...
+      # ECR_BACKEND_REPOSITORY: organic-platform-backend
+      PROJECT: wordpress-plugin
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Build the stack
+        id: build-stack
+        run: |
+          docker compose -p ${{ env.PROJECT }} --env-file .env -f docker-compose.yml up -d
```
Mwindo, line 44: lol .env won't work here, we'll need to pass in secrets and other env variables

.github/workflows/code-quality.yaml
```diff
@@ -27,3 +27,32 @@ jobs:
...
+      # ECR_BACKEND_REPOSITORY: organic-platform-backend
+      PROJECT: wordpress-plugin
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Build the stack
+        id: build-stack
+        run: |
+          docker compose -p ${{ env.PROJECT }} \
```
Mwindo, line 44: Doh! Docker isn't installed here. Note to self: reference https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - master
+    types:
+      - opened
+      - reopened
+      - synchronize
+
+jobs:
+  selenium-tests:
+    name: Selenium Tests
+    runs-on: ubuntu-latest
```
Mwindo, line 14: I'm not sure if it is worth using a self-hosted runner or not?

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,45 @@
...
+          poetry install
+
+      - name: Build the stack
+        run: |
+          mv .env.ci .env
+          poetry run ./dev.py up
+
+      - name: Run selenium tests
+        run: |
+          poetry run ./dev.py run-tests --limit_to selenium_test
```
Mwindo, line 45: Since this is not self-hosted, we shouldn't have to worry about cleanup.

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Configure poetry
+        run: |
+          poetry self add poetry-dotenv-plugin
+          poetry install
+
+      - name: Build the stack
+        run: |
+          mv .env.ci .env
+          poetry run ./dev.py up
```
Mwindo, line 41: Do we want to run the tests for all WP services? I've only done the default WP service here.

.github/workflows/integration-tests.yaml
Mwindo: Maybe a good question for @nebolsinp.

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Configure poetry
+        run: |
+          poetry self add poetry-dotenv-plugin
+          poetry install
+
+      - name: Build the stack
+        run: |
+          mv .env.ci .env
+          composer config --global github-oauth.github.com ${{ secrets.ORGANIC_DEMO_SITE_API_KEY }}
```
Mwindo, line 41: See, for instance, https://gist.github.com/nmnp/0d23e1026b54d6f5bc7caa740240a437. Without this, I was getting intermittent failures with composer installing the dependencies.

.github/workflows/integration-tests.yaml
sergunich: I think it's okay to use GH-runners for now

.github/workflows/integration-tests.yaml
sergunich: But why site API key (secrets.ORGANIC_DEMO_SITE_API_KEY) is used here?

.github/workflows/integration-tests.yaml
sergunich: Yes, best case would be to run them in parallel

sergunich (approved):

.github/workflows/integration-tests.yaml
Mwindo: 🤦
Well, maybe this didn't help after all.

.github/workflows/integration-tests.yaml
Mwindo: I will also look into [caching dependencies](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows) as well.

docker-context/Dockerfile
```diff
@@ -8,7 +8,7 @@ COPY --from=node:17.9.1 /usr/local/lib/node_modules /usr/local/lib/node_modules
 RUN ln -s ../lib/node_modules/npm/bin/npm-cli.js /usr/local/bin/npm
 RUN ln -s ../lib/node_modules/npm/bin/npx-cli.js /usr/local/bin/npx
 ARG COMPOSER_AUTH
-RUN wp --allow-root package install nlemoine/wp-cli-fixtures
+RUN wp --allow-root package install hellonico/wp-cli-fixtures
```
Mwindo, line 11: I changed this to resolve the [following issue](https://github.com/orgnc/wordpress-plugin/actions/runs/4814936863/jobs/8573141395#step:7:1724):
```
#19 [builder 10/11] RUN wp --allow-root package install nlemoine/wp-cli-fixtures
#0 1.948 Installing package hellonico/wp-cli-fixtures (dev-main)
#0 1.948 Updating /root/.wp-cli/packages/composer.json to require the package...
#0 1.949 Warning: Package name mismatch...Updating from git name 'nlemoine/wp-cli-fixtures' to composer.json name 'hellonico/wp-cli-fixtures'.
#0 1.957 Registering https://github.com/nlemoine/wp-cli-fixtures.git as a VCS repository...
#0 1.984 Using Composer to install the package...
#0 1.984 ---
#0 1.988 Loading composer repositories with package information
#0 2.617 ---
#0 2.617 Warning: Could not authenticate against github.com
#0 2.618 Error: Package installation failed.
#0 2.619 Reverted composer.json.
#19 ERROR: executor failed running [/bin/sh -c wp --allow-root package install nlemoine/wp-cli-fixtures]: exit code: 1
```

.github/workflows/integration-tests.yaml
Mwindo: I'm going to add support for wp49-php72 later:

- Tests don't currently work for WP < 5 (pre-Gutenberg). (Well, actually they fail. But it seems pointless to update the tests at the moment if there are no other tests to run.)
- There was an issue with installing the WP theme. (It's somewhere in the failed Actions--I was going to link to it, but it's getting late.)

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Install poetry
+        run: |
+          curl -sSL https://install.python-poetry.org | python3 -
+
+      - name: Configure poetry
+        run: |
+          poetry self add poetry-dotenv-plugin
+          poetry install
+
+      - name: Cache Composer dependencies
```
Mwindo, line 42: I suppose I should cache poetry deps too, which will save us maybe 10-15 seconds. I'll do this tomorrow (Thursday).

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,56 @@
...
+        uses: actions/cache@v3
+        with:
+          path: src/vendor
+          key: ${{ runner.os }}-${{ hashFiles('**/composer.lock') }}
+
+      - name: Build the stack
+        run: |
+          mv .env.ci .env
+          composer config --global github-oauth.github.com ${{ secrets.GITHUB_TOKEN }}
```
Mwindo, line 51: Not sure this is actually doing anything, but it probably doesn't hurt 🤷 .

.github/workflows/integration-tests.yaml
nebolsinp: actually self-hosted runners might save us about 2-3 minutes in each test run. because of the docker cache.
so the selenium/mysql/nginx docker layers won't be pulled every time from docker ECR, but will be immediately restored from runner's docker cache. same way as we use self-hosted runners on the platform. see the screenshots
image vs 
image

nebolsinp (approved):
looks good to me! left only one comment

.github/workflows/integration-tests.yaml
Mwindo: @nebolsinp, would the goal be to use the existing Organization k8s runners? If so, I don't think I have permissions to share those Organization runners with this repo. If not, I don't think I have permission to create new runners in AWS.

nebolsinp: No need to share runners specifically for that repo. They are shared to the entire organic github org already. just replace `runs-on: ubuntu-latest ` with  `runs-on: self-hosted` and that's it

Mwindo: I tried that but the jobs are never picked up. Maybe there is some authorization missing? In any case, I don't see the runners available in the settings [here](https://github.com/orgnc/wordpress-plugin/settings/actions/runners), which seems odd if they are shared.

nebolsinp: it's because the repository is public. and it's not recommended for self-hosted actions to work in public repositories for security [reasons](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#self-hosted-runner-security).
image

So.. I guess we have to go back to "ubuntu-latest"

nebolsinp: To be honest I completely forgot that this is a public repo :)

Mwindo: Makes sense :) Thanks Pavel!

PR #168, opened 2023-04-27
EP-5064: support for js-modules builds and code cleanup

sergunich:
- Adds support for js-module SDK builds
- Refactors how Affiliate SDK setup in admin for consistency
- Cleanup/refactor of related code
- Fixes composer deps for PHP 7.2


src/Organic/SDK/OrganicSdk.php
```diff
@@ -122,19 +122,27 @@ public function getCDNUrl() {
...
      *
-     * @return string
+     * @return array
      */
     public function getSdkV1Url() {
-        return $this->cdnUrl . 'sdk/unit-sdk.js?' . $this->siteGuid;
+        return [
+            'default' => $this->cdnUrl . 'sdk/unit-sdk.js?' . $this->siteGuid,
+            'module' => '',
+        ];
```
Mwindo, line 131: I think returning an array from `getSdkV1Url` and `getSdkV2Url` is misleading. Is there a reason we can't use a default parameter like this?
```
public function getSdkV1Url( string $type = 'default' ) {
  return $type === '' ? '' : $this->cdnUrl . 'sdk/unit-sdk.js?' . $this->siteGuid;
```

src/Organic/AdsConfig.php
```diff
@@ -52,7 +52,14 @@ public function __construct( array $rawAdsConfig, array $rawAdsRefreshRates ) {
...
-        return $this->prebidUrl;
+    public function getPrebidBuildUrl() : array {
+        $moduleSrc = '';
+        if ( preg_match( $this->MODERN_PREBID_RE, $this->prebidUrl ) ) {
+            $moduleSrc = str_replace( '.js', '.m.js', $this->prebidUrl );
+        };
+        return [
+            'default' => $this->prebidUrl,
+            'module' => $moduleSrc,
+        ];
```
Mwindo, line 63: Similar comment here to [this one](https://github.com/orgnc/wordpress-plugin/pull/168/files#r1180504311)

src/Organic/Organic.php
```diff
@@ -23,7 +23,7 @@ class Organic {
     public $version = \Organic\ORGANIC_PLUGIN_VERSION;
 
     const DEFAULT_PLATFORM_URL = 'https://app.organic.ly';
-    const DEFAULT_REST_API_URL = 'https://api.organiccdn.io/';
+    const DEFAULT_REST_API_URL = 'https://api.organiccdn.io';
```
Mwindo, line 26: 👍

src/Organic/PageInjection.php
```diff
@@ -138,37 +138,76 @@ function ( $content ) {
         );
     }
 
+    public function injectBrowserSDKInAdmin() {
```
Mwindo, line 141: Much nicer. Thanks!

src/Organic/PageInjection.php
```diff
@@ -487,12 +537,39 @@ function loadScript(id, src) {
...
+            wp_print_script_tag(
+                [
+                    'id' => $id,
+                    'src' => $src,
+                    'async' => true,
+                ]
+            );
+            return;
+        }
+        wp_print_script_tag(
```
Mwindo, line 555: It might be helpful to clarify why both scripts need to be added in this case, e.g.:
```
// The script to be used if modules are supported.
wp_print_script_tag(...
);
// The script to be used if modules are not supported.
wp_print_script_tag(...
);
```

Mwindo (approved):
I left a couple of small comments, but LGTM, and Affiliate seems to be running correctly. Thanks, @sergunich!

sergunich:
Thanks, @Mwindo! I'll merge PR as-is because want to release it before vacation but will get back to your suggestions after 👍

Mwindo:
> Thanks, @Mwindo! I'll merge PR as-is because want to release it before vacation but will get back to your suggestions after 👍

Sounds good! Have a great time :)

PR #175, opened 2023-05-18
EP-5546: Widget Insertion Refresh Part 2

Mwindo:
Somehow the `pageshow` event is being triggered too early in some cases; setting a higher timeout fixed the issue, but that's obviously a hack.

I feel doubtful about finding the "right" event to guarantee that WordPress has fully "loaded" if `pageshow` fails; instead, it feels safer to reference WordPress functionality directly. This alternative solution is based on https://gist.github.com/KevinBatdorf/fca19e1f3b749b5c57db8158f4850eff. (https://github.com/WordPress/gutenberg/issues/8379 was also helpful.)

https://github.com/orgnc/wordpress-plugin/assets/47676832/a43af454-3578-426a-96ff-b11d4e3abaee

### DIFF
--- a/src/blocks/affiliate/shared/helpers.js
+++ b/src/blocks/affiliate/shared/helpers.js
@@ -1,5 +1,7 @@
 export const refreshAffiliateWidgets = () => {
-  window.organic ||= {};
-  window.organic.cmd ||= [];
-  window.organic.cmd.push(({ affiliate }) => affiliate?.processPage());
+  setTimeout(() => {
+    window.organic ||= {};
+    window.organic.cmd ||= [];
+    window.organic.cmd.push(({ affiliate }) => affiliate?.processPage());
+  }, 0);
 };
--- a/src/blocks/affiliate/shared/helpers.js
+++ b/src/blocks/affiliate/shared/helpers.js
@@ -4,9 +4,8 @@ import { select, subscribe } from '@wordpress/data';
 async function whenEditorIsReady() {
   return new Promise((resolve) => {
     const unsubscribe = subscribe(() => {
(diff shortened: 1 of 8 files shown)


src/blocks/affiliate/shared/helpers.js
```diff
@@ -1,5 +1,36 @@
...
+  }, 0);
+};
+
+export const refreshAffiliateWidgetsOnEdit = () => {
+  refreshAffiliateWidgets();
+};
+
+export const refreshAffiliateWidgetsOnSave = () => {
+  whenEditorIsReady().then(() => {
+    if (!window.wp_organic_affiliate_processed) {
```
Mwindo, line 31: I can add a comment, something like, "Prevent multiple calls to pageProcess."

Mwindo: (Since every widget on the admin page will call this.)

sergunich: Adding a comment sounds like a good idea; another suggestion is to use `window.__wpOrganicAffiliateProcessed` flag for consistency

sergunich (approved):
👍

Mwindo:
> The fix for refreshing the widgets on page load works in Chrome/Edge, but it doesn't fix the issue in Safari/Firefox :/ So more to come.

False alarm. After a series of tests, I've confirmed this had to do with page caching. That said, I've refactored out the affiliate toolbar (should have been like this anyway) and added a "Refresh display" option as a last resort.

Mwindo:
Toolbar looks like this now:
Screen Shot 2023-05-22 at 3 11 56 PM

PR #183, opened 2023-06-29
EP-4812: Refactor to Add Ads Enabled Option

Mwindo:


src/Organic/AdminSettings.php
```diff
@@ -77,11 +77,6 @@ protected function updateSettings() {
         }
 
         // Organic Settings
-        $this->organic->updateOption(
-            'organic::enabled',
-            isset( $_POST['organic_enabled'] ) ? true : false,
-            false
-        );
```
Mwindo, line 84: "Enabled" will be offloaded to WordPress, i.e., plugin is activated or deactivated. If activated, we will now have more granular options about what, exactly, is enabled, e.g., Ads, Affiliate, etc.

src/Organic/AdConfigSyncCommand.php
```diff
@@ -49,8 +49,11 @@ public function run() {
      * @return void
      */
     public function __invoke( $args ) {
-        // Only both trying if the API key is set
-        if ( ! $this->organic->isEnabledAndConfigured() ) {
+        // Only worth trying if Ads is enabled and the API key is set
```
Mwindo, line 52: I think Ads needs to be enabled, but maybe not?

Mwindo: (E.g., maybe similar to Ads.txt redirect not needing Ads enabled)

src/Organic/GraphQL.php
```diff
@@ -144,7 +144,7 @@ public function register_organic_types() {
                     $adsConfig = $organic->getAdsConfig();
                     $testEnabled = $organic->useSplitTest();
                     return [
-                        'organicEnabled' => $organic->isEnabledAndConfigured(),
+                        'organicEnabled' => $organic->isConfigured(),
```
sergunich, line 147: Right now, a lot of Recurrent sites do have Organic WP plugin enabled and configured in WP, but the "integration" checkbox is disabled

That's because Organic Web has a hard dependency on Organic WP plugin GQL API and they can't just disable/delete the whole plugin

And by passing the `$organic->isConfigured()` as `organicEnabled` value will lead to injecting and setting up Organic SDK on those sites (even if Ads and Affiliate are disabled the Core is always enabled in the SDK)
https://github.com/orgnc/organic-content/blob/ae0c8844ae433483ac26300a8f8a2b8489490a19/frontend/packages/core/components/common/OrganicSDK/DocumentHeadOrganicTags.tsx#L184

src/Organic/Organic.php
```diff
@@ -399,14 +422,14 @@ public function setPostTypes( $types ) {
      * @return bool
      */
     public function useCmpBuiltIn() : bool {
-        return $this->isEnabled() && $this->cmp == 'built-in';
+        return $this->adsEnabled() && $this->cmp == 'built-in';
```
sergunich, line 425: CMP can be used without Ads

src/Organic/Organic.php
```diff
@@ -447,11 +474,11 @@ public function useAffiliate() : bool {
      * @return bool
      */
     public function useAmp() : bool {
-        return $this->isEnabledAndConfigured() && $this->ampEnabled;
+        return $this->useAds() && $this->ampEnabled;
```
sergunich, line 477: Affiliate has AMP integration too
Overall I considered AMP and Prefill an plugin-level features, not something related to Ads or Affiliate

src/Organic/AdminSettings.php
```diff
@@ -190,6 +190,8 @@ public function showSettings() {
         $ad_slots_prefill_enabled = $this->organic->getOption( 'organic::ad_slots_prefill_enabled' );
         $log_to_sentry = $this->organic->getOption( 'organic::log_to_sentry', true );
 
+        $ads_enabled = $this->organic->getOption( 'organic::ads_enabled' );
```
sergunich, line 193: This means that after the plugin upgrade, the Ads will become disabled because the `organic::ads_enabled` field in DB will be empty

If the `organic::ads_enabled` is empty, we should pre-populate it based on behavior of previous versions of the plugin: if plugin/integration is enabled and configured, then set it to `true`

src/Organic/AdConfigSyncCommand.php
sergunich: Yeah, I think it worth syncing to make sure we have config before enabling Ads

src/Organic/GraphQL.php
Mwindo: > That's because Organic Web has a hard dependency on Organic WP plugin GQL API

Ah, ok, I see. I wonder if we can modify the language or add plugin field descriptions to make this all clearer? As a client looking at the settings page, I would be extremely confused. What does it mean for the integration to be enabled--isn't the plugin already active? If it's enabled but Ads, Affiliate, and Campaigns are not, then what is the plugin doing? Etc.

Tagging @cjkie here too.

src/Organic/AdminSettings.php
Mwindo: Good call, thank you!

src/Organic/AdminSettings.php
Mwindo: Nevermind, see https://github.com/orgnc/wordpress-plugin/pull/183#discussion_r1254378768

src/Organic/Organic.php
Mwindo: When is prefill used outside of Ads?

PR #188, opened 2023-07-11
EP-5180 / EP-5622 :: Additional data for Organic and resync handling

rrdial:


src/Organic/Organic.php
```diff
@@ -1071,6 +1074,28 @@ public function fullResyncContent( $batch = 50, $offset = 0, $sleep_between = 0
         return $updated;
     }
 
+    public function getContentResyncStartedAt(): DateTimeImmutable {
+        return DateTimeImmutable::createFromFormat(
+            DATE_ATOM,
+            $this->getOption(
+                'organic::content_resync_started_at',
+                '2003-05-27T00:00:00+00:00'
```
rrdial, line 1082: Just a fun Easter Egg since there needed to be a default date sometime in the past. https://wordpress.org/news/2003/05/wordpress-now-available/

Mwindo: I was going to ask about this particular default :)

src/Organic/Organic.php
```diff
@@ -1312,6 +1337,16 @@ public function syncPluginConfig() {
...
             }
+            if ( $config['triggerContentResync'] && ! $this->contentResyncTriggeredRecently() ) {
+                global $wpdb;
+                $wpdb->get_results(
+                    $wpdb->prepare(
+                        "UPDATE $wpdb->postmeta SET meta_value = 'unsynced' WHERE meta_key = %s",
+                        SYNC_META_KEY
+                    )
+                );
+                $this->updateContentResyncStartedAt();
```
Mwindo, line 1348: I wonder if these lines should be pulled out into separate function. Not really dead set on this.

Mwindo (approved):
One thing I don't love is that the value we get in Django for `content_resync_started_at` won't really reflect when the content "actually" started to sync; it looks like that could take up to an hour (depending on when the command is invoked). But I don't see a great way to avoid this--maybe it's worth having a `content_resync_finished_at` option?

Anyways, LGTM overall. Thanks @rrdial!

src/Organic/Organic.php
rrdial: Updated to include the appropriate time. Accurate data is very important.

tonystanhope (commented):
👍

PR #192, opened 2023-07-18
Bugfix: config sync content sync

rrdial:
The primary issue fixed here is the OrganicSDK. We were previously missing an intermediary array key for `syncWordpressPluginConfig` which is now included.

Through the course of debugging this issue, I added some minor code restructuring that I'm leaving in this PR with other minor fixes/optimizations.

### DIFF
--- a/src/Organic/AdminSettings.php
+++ b/src/Organic/AdminSettings.php
@@ -315,7 +315,7 @@ class AdminSettings {
                             name="organic_amp_ads_enabled"
                             <?php echo $amp_ads_enabled ? 'checked' : ''; ?>
                         />
-                        AMP Intergation Enabled
+                        AMP Integration Enabled
                     </label>
                 </p>
                 <p>
--- a/src/Organic/AdminSettings.php
+++ b/src/Organic/AdminSettings.php
@@ -470,7 +470,7 @@ class AdminSettings {
                 <?php $this->injectEnvInfo( 'PLATFORM_URL', $this->organic->getPlatformUrl() ); ?>
                 <?php $this->injectEnvInfo( 'ADS_TXT_URL', $this->organic->getAdsTxtManager()->getAdsTxtUrl() ); ?>
                 <?php $this->injectEnvInfo( 'Next Content Sync', $contentSyncCron ? $contentSyncCron->format( DATE_ATOM ) : 'false' ); ?>
-                <?php $this->injectEnvInfo( 'Content Re-Sync Triggered At', $this->organic->getContentResyncStartedAt() ); ?>
+                <?php $this->injectEnvInfo( 'Content Re-Sync Triggered At', $this->organic->getContentResyncStartedAt()->format( DATE_ATOM ) ); ?>
                 <?php if ( ! $ads_txt_redirect ) { ?>
                     <p>
                         <label>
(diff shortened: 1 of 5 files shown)


src/Organic/Organic.php
```diff
@@ -198,7 +198,7 @@ public function getOption( $name, $default = false ) {
             $result = get_option( $name );
 
             // Fallback to old version if it exists instead
-            if ( ! $result ) {
+            if ( false === $result ) {
```
Mwindo, line 201: 👍

src/Organic/Organic.php
```diff
@@ -1337,15 +1351,8 @@ public function syncPluginConfig() {
...
-                    $wpdb->prepare(
-                        "UPDATE $wpdb->postmeta SET meta_value = 'unsynced' WHERE meta_key = %s",
-                        SYNC_META_KEY
-                    )
-                );
-                $this->updateContentResyncStartedAt();
+            if ( true === $config['triggerContentResync'] ) {
+                $this->triggerContentResync();
             }
         }
```
Mwindo, line 1357: Can we move the `if (! empty( $config ) ) { ...` up through these lines into the above try block?

Mwindo (approved):
One small comment, but LGTM. Thanks @rrdial!

src/Organic/Organic.php
rrdial: Updated. Not sure if this is exactly what you had in mind.

rrdial: Updated!

src/Organic/helpers.php
```diff
@@ -25,3 +27,40 @@ function is_valid_uuid( string $uuid ) {
...
+/**
+ * Copy of the built-in WordPress function `get_edit_post_link`.
+ * Retrieves the edit post link for post without a permissions check.
+ *
+ * @param int|WP_Post $id      Optional. Post ID or post object. Default is the global `$post`.
+ * @param string      $context Optional. How to output the '&' character. Default '&amp;'.
+ * @return string|null|void    The edit post link for the given post. Null if the post type does not
+ *                             exist or does not allow an editing UI.
+ */
+function get_edit_post_link( $id = 0, string $context = 'display' ) {
```
Mwindo, line 40: Ah, so `editUrl` was required but we couldn't get it because of permissions issue. Should we explicitly import this function in `OrganicSdk.php` just to make it clear we are using our custom version?

src/Organic/AdminSettings.php
```diff
@@ -313,7 +315,7 @@ class="organic-wide-input"
                             name="organic_amp_ads_enabled"
                             <?php echo $amp_ads_enabled ? 'checked' : ''; ?>
                         />
-                        AMP Intergation Enabled
+                        AMP Integration Enabled
```
Mwindo, line 318: Good catch!

Mwindo (approved):
One small comment again, but otherwise looks good. I like the added debug lines :) Thanks @rrdial!

src/Organic/Organic.php
```diff
@@ -814,7 +813,7 @@ public function syncPost( WP_Post $post ) {
...
 
         $canonical = get_permalink( $post );
-        $edit_url = get_edit_post_link( $post );
+        $edit_url = \Organic\get_edit_post_link( $post );
```
Mwindo, line 816: 🎉

src/Organic/helpers.php
rrdial: Meh. Really it would just be redundant because we're in the same namespace. That said, for an improved developer experience, I'll make it explicit. Not an import, but a fully qualified function call so there won't be any _human_ ambiguity.

Mwindo: Even better!

PR #203, opened 2023-10-16
EP-6275 :: Bugfix - sync processed content, not the raw content

rrdial:


jdemaris (approved):
Looks good. The `get_post_field` function is old and established, so I don't foresee any version compatibility issues.

ZLevine (approved):
LGTM.php

PR #204, opened 2023-10-19
EP-6275 :: Conditionally trigger content sync on plugin update

rrdial:


src/organic.php
```diff
@@ -45,6 +45,15 @@ function init_organic_sentry( string $dsn, string $environment ) : ?\Sentry\Stat
     getenv( 'ORGANIC_CDN_URL' ) ?: getenv( 'EMPIRE_CDN_URL' )
 );
 
+// On update or activation, we need to make sure content is re-synced because of changes to the plugin code
+register_activation_hook(__FILE__, function () use ($organic) {
+    $resynced_on_version = get_option( 'organic::resynced_on_version', '0.0.0' );
+    if ( version_compare( '1.14.4', $resynced_on_version, 'gt' ) ) {
```
jdemaris, line 51: Any reason we wouldn't do this for backwards versions as well? As in just check if they don't match and force a resync if they don't match?

jdemaris: Also - why hardcode the magic number? Why not compare to `Organic\ORGANIC_PLUGIN_VERSION`?

rrdial: So, this is a continuation of the [previous change](https://github.com/orgnc/wordpress-plugin/pull/203). I'm looking for a way to trigger a re-sync because of a bugfix that was implemented in (should be) version `1.14.1`. Prior to that, we would still be getting bad data. But if a client updates the plugin at their own pace and doesn't _actual_ do the update until, for example, version `1.16.0`, then we would still want to trigger the re-sync. _However_, we don't want to re-sync _every_ time the plugin is updated. Also, `Organic\ORGANIC_PLUGIN_VERSION` is replaced during the CI/CD build with the actual release number, so we would, again, essentially be triggering a re-sync with every update.

jdemaris: Got it - okay that makes sense and not a blocker. Only remaining fixes are the codestyle / build issues

PR #208, opened 2024-04-12
Remove obsolete messaging from WP settings

ZLevine:
This PR removes the "WARNING: DO NOT USE IN PRODUCTION" message that's very, very obsolete.

image


jdemaris (approved):
Finally! Thank you!

